PR #7, opened 2024-06-17
Add script for retrieving sequences for EST using the Family option.

### LINKED ISSUE #6: Refactor sequence import
Sequences are imported in various ways, and each method currently has its own script. Refactor these scripts into one coherent command line application. Fork the nextflow-test branch and place the script in `src/est/import/`.

nilsoberg:


1ndy:
A few things:
- Would it make more sense to use the `lib/` that is in the project root? The SSN pipeline also lives in this repo so if there is opportunity for code reuse I would rather have it there.
- Can you add some print statements to describe what the program is doing? Just high level things like "Computing accession IDs" or "Saving \<data\> to \<file\>" so that debugging is easier.
- Can you make fasta.metadata tab separated? It looks like it could just be two columns, seqid and source (and possibly a third "length" column if you plan to add that as the comments suggest)
- Is there a reason not to add headers to all tab-separated outputs? This would make it easier to understand what data is stored in the file.
- We should discuss where each of the files other than `allsequences.fa` are used in the pipeline. I have been computing some of the same information (such as sequence lengths) in later stages.
- Would it make sense to split the program into a few pieces? One to fetch accession IDs, one to translate the IDs to sequences
- I would prefer not to use any environment variables and instead just centralize configuration in a file unless there is a good reason to mix the two

1ndy:
Also, can `--output-dir` be made optional and default to writing to the current directory

PR #26, opened 2024-07-09
10 add accession import option

### LINKED ISSUE #10: Add Accession Import Option

nilsoberg:


src/est/import/lib/EFI/Import/Source/Accession.pm
```diff
@@ -0,0 +1,158 @@
...
+
+####################################################################################################
+# 
+#
+
+
+sub parseAccessions {
+    my $self = shift;
+
+    $self->{logger}->message("Parsing accession file $self->{acc_file}");
```
1ndy: This line (and a similar call on line 120) cause a crash:
```
Using accessions as source
Retrieving accession IDs from source efi_db.sqlite
Can't call method "message" on an undefined value at src/est/import/lib/EFI/Import/Source/../../../EFI/Import/Source/Accession.pm line 82.
```
Looks like the logger isn't initialized?

nilsoberg: Both of those logger statements can/should be removed.

Issue #20, opened 2024-07-08
Write basic tests


1ndy:
Write basic tests that test the creation of parameter files under different import modes. These should not be designed to test edge cases but rather confirm that no changes have caused regressions.

- [ ] Add option to specify which config to use

nilsoberg:
`--blast-query smalldata/blast_query.fa` can be added to `create_est_nextflow_params.py` line in `est_sequence_blast.sh` to support BLAST input option.

A sequence for testing can be extracted from the small dataset:

`perl -e '$h=0; while (<>) { last if (m/^>/ and ++$h > 1); next if m/^>/; $seq .= $_; } print $seq;' < smalldata/test.fasta > smalldata/blast_query.fa`

PR #27, opened 2024-07-11
11 add sequence blast import option

### LINKED ISSUE #11: Add Sequence BLAST Import Option

nilsoberg:
To get test data, run something like:

`
perl -e '$h=0; while (<>) { last if (m/^>/ and ++$h > 1); next if m/^>/; $seq .= $_; } print $seq;' < smalldata/test.fasta > smalldata/blast_query.fa
`

Also, add `--blast-query smalldata/blast_query.fa` to `create_est_nextflow_params.py` line in `est_sequence_blast.sh` in #20 branch.


create_est_nextflow_params.py
```diff
@@ -21,19 +21,20 @@ def add_args(parser: argparse.ArgumentParser):
     common_parser.add_argument("--efi-db", required=True, type=str, help="Name of the MySQL database to use (e.g. efi_202406) or name of the SQLite file")
     common_parser.add_argument("--multiplex", action="store_true", help="Use CD-HIT to reduce the number of sequences used in analysis")
     common_parser.add_argument("--blast-evalue", default="1e-5", help="Cutoff E value to use in all-by-all BLAST")
+    common_parser.add_argument("--family-id-format", type=str, required=False, default="UniProt", choices=["UniProt", "UniRef90", "UniRef50"])
```
1ndy: Since this is no longer family-specific, lets change it to `--sequence-version` as in the Perl script.

Also, if the `default` kwarg is set, `require=False` is redundant.

create_est_nextflow_params.py
```diff
@@ -82,6 +83,12 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
             fail = True
         else:
             args.accessions_file = os.path.abspath(args.accessions_file)
+    elif args.import_mode == "blast":
+        if not os.path.exists(args.blast_query_file):
+            print(f"BLAST query file '{args.blast_query_file}' does not exist")
+            fail = True
+        else:
+            args.blast_query_file = os.path.abspath(args.blast_query_file)
```
1ndy: Code looks good but lets make this the `if` so that the order is the same everywhere in the file:
- blast
- family
- fasta
- accessions

create_est_nextflow_params.py
```diff
@@ -122,12 +130,12 @@ def render_params(output_dir, duckdb_memory_limit, duckdb_threads, fasta_shards,
         "import_mode": import_mode,
         "exclude_fragments": exclude_fragments,
         "multiplex": multiplex,
-        "blast_evalue": blast_evalue
+        "blast_evalue": blast_evalue,
+        "family_id_format": family_id_format
```
1ndy: Change to `"sequence_version": sequence_version` as described above

create_est_nextflow_params.py
```diff
@@ -137,6 +145,10 @@ def render_params(output_dir, duckdb_memory_limit, duckdb_threads, fasta_shards,
         params |= {
             "accessions_file": accessions_file
         }
+    elif import_mode == "blast":
+        params |= {
+            "blast_query_file": blast_query_file
+        }
```
1ndy: Make this the first clause of the if statement as above

est.nf
```diff
@@ -4,23 +4,37 @@ process get_sequence_ids {
         path 'import_stats.json', emit: 'import_stats'
         path 'sequence_metadata.tab', emit: 'sequence_metadata'
         path 'sunburst_ids.tab', emit: 'sunburst_ids'
+        path 'blast_hits.tab', emit: 'blast_hits'
```
1ndy: If this will only be present when the blast option runs, it needs to be marked optional
```
path 'blast_hits.tab', emit: 'blast_hits', optional: true
```

We might also consider executing `cp` within the code block but I'm not sure I want to have side-effects like that

est.nf
```diff
@@ -4,23 +4,37 @@ process get_sequence_ids {
...
+    if (params.import_mode == "family") {
         """
-        perl $projectDir/src/est/import/get_sequence_ids.pl $common_args --family ${params.families} --sequence-version ${params.family_id_format}
+        perl $projectDir/src/est/import/get_sequence_ids.pl $common_args --family ${params.families}
         """
-    else if (params.import_mode == "accessions") {
+    } else if (params.import_mode == "accessions") {
         """
-        perl $projectDir/src/est/import/get_sequence_ids.pl $common_args --accessions ${params.accessions_file}
+        perl $projectDir/src/est/import/get_sequence_ids.pl $common_args --user-accession-file ${params.accessions_file}
```
1ndy: Its not clear how this is related to the blast import option

est.nf
```diff
@@ -4,23 +4,37 @@ process get_sequence_ids {
...
         """
-    }
-    else
+    } else if (params.import_mode == "blast") {
+        // blast_hits.tab is provided as an output to the user
+        """
+        blastall -p blastp -i ${params.blast_query_file} -d ${params.fasta_db} -m 8 -e ${params.blast_evalue} -b ${params.num_blast_matches} -o init.blast
+        SZ=\$(wc -c "init.blast" | cut -d' ' -f1)
+        if [[ \$SZ != 0 ]]; then
```
1ndy: Any reason not to use `[ -s init.blast ]`? Or `stat --printf="%s"` instead of the pipe

est.nf
```diff
@@ -4,23 +4,37 @@ process get_sequence_ids {
...
+            cat init.blast | grep -v '#' | cut -f 1,2,3,4,12 | sort -k5,5nr > init_blast.tab
+            cat init.blast | grep -v '#' | cut -f 2,11 | sort -k2nr > blast_hits.tab
```
1ndy: Can we pass the file straight into `grep` here to have one less pipe?

Also, I remember `sort`ing requiring a lot of memory in old BLASTreduce, do these files need to be sorted? The import process is only assigned 32GB of memory by default.

If we do need to sort, can we define a total order (use all columns)?

est.nf
```diff
@@ -4,23 +4,37 @@ process get_sequence_ids {
...
+            perl $projectDir/src/est/import/get_sequence_ids.pl $common_args --blast-output init_blast.tab --blast-query ${params.blast_query_file}
+        else
+            echo "BLAST Failed. Check input sequence."
```
1ndy: A more helpful error message might mention that BLAST did not return any matches

est.nf
```diff
@@ -71,7 +90,7 @@ process import_fasta {
 
     """
     # produces a mapping.txt file
-    perl $projectDir/src/est/import/get_sequence_ids.pl --efi-config ${params.efi_config} --efi-db ${params.efi_db} --mode fasta --fasta ${params.uploaded_fasta_file}
+    perl $projectDir/src/est/import/get_sequence_ids.pl --efi-config ${params.efi_config} --efi-db ${params.efi_db} --mode fasta --user-fasta-file ${params.uploaded_fasta_file}
```
1ndy: Its not clear how this is related to the blast import option. This is a topic for another branch, but why does `import_fasta.pl` have `--uploaded-fasta` while `get_sequence_ids.pl` has `--user-fasta-file`? They use the same file so I think we should reconcile the option name.

src/est/import/append_blast_query.pl
1ndy, line 1: Could we accomplish the same thing with `cat` or shell scripting?

src/est/import/lib/EFI/Import/Metadata.pm
```diff
@@ -24,11 +25,13 @@ use constant FIELD_UNIREF50_CLUSTER_SIZE => "UniRef50_Cluster_Size";
 use constant FIELD_UNIREF90_CLUSTER_SIZE => "UniRef90_Cluster_Size";
 use constant FIELD_UNIREF100_CLUSTER_SIZE => "UniRef100_Cluster_Size";
 
+use constant INPUT_SEQ_ID => "zINPUTSEQ";
```
1ndy: I think BLAST uppercases all accession IDs, lets make this uppercase as well so that we don't have the possibility of losing a reference somewhere. This caused the pipeline to fail because some alignment scores were null.

est.nf
```diff
@@ -58,6 +72,11 @@ process cat_fasta_files {
     """
     cat $input > all_sequences.fasta
     """
+    if (params.import_mode == "blast") {
+        """
+        perl $projectDir/src/est/import/append_blast_query.pl --blast-query-file ${params.blast_query_file} --output-sequence-file all_sequences.fasta
+        """
+    }
```
1ndy: This won't work the way I think you intended it to. A nextflow process [is a Groovy function that returns a string](https://www.nextflow.io/docs/latest/process.html#conditional-scripts). The last thing on the stack here could be nothing if the condition evaluates to false. You need to build a string here instead.
```
    cat_cmd = "cat $input > all_sequences.fasta"
    if (params.import_mode == "blast") {
        """
        $cat_cmd
        perl $projectDir/src/est/import/append_blast_query.pl --blast-query-file ${params.blast_query_file} --output-sequence-file all_sequences.fasta
        """
    } else {
        cat_cmd
    }
```

est.nf
```diff
@@ -231,6 +250,9 @@ workflow {
         sequence_id_files = fasta_import_files
     } else {
         sequence_id_files = get_sequence_ids()
+
+        // split up the sequence ID list into separate files so we can parallelize the
+        // process of retrieving sequences from the BLAST sequence database
```
1ndy: This is admittedly very pedantic but we are not _parallelizing_ so much as we are _making it parallelizable_. We should get in the habit of thinking that nextflow decides how to paralleize it, we just define the dependency graph. It would be more accurate to say that we split the file to enable parallel sequence retrieval

src/est/import/README.md
1ndy, line 1: New documentation should be added under `docs/`

est.nf
nilsoberg: This is only present for the blast option.

est.nf
nilsoberg: I will move these and related changes to a separate branch

src/est/import/append_blast_query.pl
nilsoberg: Implementing this through `cat`, etc, was my initial thought.  I was wanting to avoid a hard-coded value, and the `EFI::Import::Metadata` module defines `INPUT_SEQ_ID` constant.  Open to other approaches.

src/est/import/lib/EFI/Import/Metadata.pm
nilsoberg: How does it work in the FASTA option, where the sequences are prefixed with lowercase 'zzz's?

est.nf
nilsoberg: My understanding is that `-s` is a bash extension specific to Linux.  `stat` is likewise Linux-specific.  Those may or may not be present in OS X, for example.  This may be a non-issue.

est.nf
nilsoberg: The entire first line can be replaced by the `get_sequence_ids.pl` script since column extraction and sorting can happen internally to the script.  The second line can be replaced with something like `awk {...} init.blast | sort -k2nr > blast_hits.tab`, replacing two pipes, `cat`, `grep`, and `cut` with `awk`.  The sort memory point you brought up is probably not an issue since the file that is being sorted is ~1000 lines long.

est.nf
1ndy: The process script is interpreted by Nextflow as a Bash script by default
https://www.nextflow.io/docs/latest/process.html#scripts-a-la-carte

The docker image is meant to be used if someone is running this locally so their underlying platform should not matter.

est.nf
1ndy: That sounds perfect

src/est/import/append_blast_query.pl
1ndy: That makes sense, we can keep this approach

src/est/import/lib/EFI/Import/Metadata.pm
1ndy: I'm not sure. I saw this discrepancy between the transcoded FASTA (zINPUTSEQ) file and the output of the axa_blast stage (ZINPUTSEQ). I'll trace it through a little more and see whats going on

1ndy: The testing we have done on FASTA files has only included sequences which are in our test database. I think this means that the accession IDs are not rewritten and so that is why we haven't seen this problem yet.

UniProt [defines their accession ID format](https://www.uniprot.org/help/accession_numbers) to include only uppercase letters. I think we should follow that convention when we need to rewrite sequences, because it seems like BLAST does. I will add a test that uses a FASTA file with unknown Accession IDs to catch this in the future.

nilsoberg: Unknown sequences were identified by a lowercase 'z' at the start of the ID.  If it's uppercase then the unknown sequence ID checks need to be updated everywhere.  UniProt includes sequences that start with a Z, but we can still identify unknown sequences by checking for consecutive ZZs, since UniProt is always A-Z at the start followed by a number.

nilsoberg: I'll create an issue for this change.

1ndy: Can we avoid storing information about UniProt membership in the sequence altogether?

nilsoberg: The FASTA header/metadata file can store any type of alphanumeric identifier.  Downstream tools such as the GNT, GND, and Cluster Analysis tools require UniProt IDs to function, so the assumption throughout the scripts is that the IDs are UniProt.  `zzz` IDs are excluded in GND, GND, CA, etc.  I've been wanting to change the scripts to work with any kind of ID because that will become important for our projects with external partners.  Changes to the ID format/assumptions will also require changes to the database.

PR #38, opened 2024-07-16
15 pod in sphinx

### LINKED ISSUE #15: pod in sphinx
Come up with a way to use perlpod in the Perl portions of the pipeline to get documentation on Perl functions into sphinx.

1ndy:


Makefile
```diff
@@ -14,4 +14,7 @@ docs-spelling:
 	sphinx-build -b spelling docs/ build/
 
 docs-clean:
-	rm -rf build/
\ No newline at end of file
+	rm -rf build/
+
+docs-perlpod:
```
nilsoberg, line 25: Should this also include `*.pm` files?

nilsoberg (approved):

Makefile
nilsoberg: No - because the documentation we provide describes scripts that users interact with.  Modules will be documented internally.

1ndy: We should render documentation from Perl modules, this process does not account for them right now. I had not thought about it. If we include `*.pm` in this target, I'm worried that the pipeline documentation will get crowded. Should we try to develop a separation between docs for scripts that run in the pipeline and libraries we have written that are used by the scripts?

nilsoberg: I thought about it, and I think for now we should only include `*.pl`, unless there is an API that is designed to be used external to the scripts.  The modules can be reused as-is, but there is currently no need for that.

PR #54, opened 2024-07-29
48 incrementally publish files

### LINKED ISSUE #48: Incrementally publish files
The Nextflow pipelines currently have a `finalize_output` stage. This was intended to be used to compress the final output files. Output finalization may be platform specific and therefore should be handled by a script which runs after the pipeline succeeds. Remove the `finalize_output` stages from the nextflow pipelines and replace them `publishDir` directives within processes.

1ndy:


nilsoberg (approved):
I like this change. It makes things clearer.

PR #58, opened 2024-08-02
52 refactor common functionality into lib

### LINKED ISSUE #52: Refactor common functionality into lib
Some functionality has been duplicated in the pipeline scripts. In particular,

- [ ] Parquet transcoding functions have been copied several times
- [ ] SQL template rendering exists in several places

Refactor this functionality into a a module and explore moving it to the `lib/` directory. Search for other duplicated functionality and refactor it as well.

1ndy:
There is a lot here. A lot of it consists of moving files around but one notable addition is `lib/pyEFI`, a python package which encapsulates common functionality used across packages. Right now it contains code for transcoding and rendering SQL templates (and a few other things) but my hope is that it can continue to grow as we work through the larger nextflow refactor.

Adding the package introduces a build/install step the needs to be executed if the package source is updated. Downstream wrappers (such as KBase) might need to adjust their environments to accommodate this.


Dockerfile
```diff
@@ -23,8 +24,9 @@ RUN curl -L -o /opt/cd-hit.tar.gz https://github.com/weizhongli/cdhit/releases/d
...
+# set up Perl environment
+RUN apt update && apt install -y cpanminus libdbd-mysql-perl zip && cpanm --installdeps .
+
 # set up python environment
 RUN pip3 install -r app/requirements.txt
-
-# set up Perl environment
-RUN apt update && apt install -y cpanminus libdbd-mysql-perl zip && cpanm --installdeps .
\ No newline at end of file
+RUN pip3 install opt/pyEFI
```
nilsoberg: Should this be `/opt/pyEFI` not `opt/pyEFI`?

nilsoberg (commented):
I have a comment about `Dockerfile` but other than that this PR looks good.

PR #66, opened 2024-08-06
59 tests require efi env

### LINKED ISSUE #59: Tests require efi-env
When attempting to run the included tests via the default docker configuration, I discovered that they would not 
execute unless the `efi-env` environment folder was present and could be sourced. I created this environment but did 
not install any package in it; this allowed the tests to run.

The parameter generation scripts (with the exception of the `bin/create_nextflow_job.py`) are designed to run using only the Python standard library and should not require activating any environment.

Remove environment activation from `tests/runtests.sh`. If it is needed to test a module, it may be activated there.

1ndy:


nilsoberg (approved):
Based on the update, I assume that the `.py` scripts are executable.  Is that correct?

1ndy:
Yes, all are executable

PR #78, opened 2024-08-22
28 implement uniref support in import process

### LINKED ISSUE #28: Implement UniRef support in import process
All import options should be able to import UniRefXX sequences in addition to UniProt.

nilsoberg:
Tested for UniProt, UniRef90, and UniRef50.  Added documentation to the modules that were updated.  Extensive refactors to better support UniRef metadata.

### DIFF
--- a/bin/create_est_nextflow_params.py
+++ b/bin/create_est_nextflow_params.py
@@ -79,11 +79,16 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
             fail = True
         else:
             args.blast_query_file = os.path.abspath(args.blast_query_file)
-        if args.sequence_version != "uniprot" and len(args.import_blast_fasta_db) == 0:
-            print(f"BLAST FASTA db file '{args.import_blast_fasta_db}' for importing not found")
-            fail = True
+        if args.sequence_version != "uniprot":
+            if args.import_blast_fasta_db is None:
+                print("--import-blast-fasta-db is required when sequence version is not uniprot")
+                fail = True
+            else:
+                # Use the UniRef database for the BLAST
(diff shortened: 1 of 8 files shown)


bin/create_est_nextflow_params.py
```diff
@@ -78,6 +79,11 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
             fail = True
         else:
             args.blast_query_file = os.path.abspath(args.blast_query_file)
+        if args.sequence_version != "uniprot" and len(args.import_blast_fasta_db) == 0:
+            print(f"BLAST FASTA db file '{args.import_blast_fasta_db}' for importing not found")
```
1ndy: `argparse` will initialize optional parameters to `None` if they are not included on the commandline and `len(None)` results in an error. Also, telling the user that the file was not found seems less descriptive than telling them that they must include it since they are not using uniprot sequences. 

```python
if args.sequence_version != "uniprot" and args.import_blast_fasta_db is None:
    print(f"--import-blast-fasta-db is required when sequence version is not uniprot")
else:
    if not os.path.exists(args.import_blast_fasta_db):
        print(f"BLAST FASTA db file '{args.import_blast_fasta_db}' does not exist"
        fail = True
```
Finally, it is fine to do the `abspath` in the `else`, but other arguments are reassigned at the bottom of the function. Let's try to be consistent about where we do this operation.

nilsoberg: Because of a dependency on `args.fasta_db`, I elected to keep everything in the else.  The assignments at the bottom logically go there so I'm keeping them there for the time being.

PR #79, opened 2024-08-23
Rename command line args for sequence import for clarity

### LINKED ISSUE #32: Clarify import process command line arguments
Clarify the command line arguments for FASTA and Accession import processes.

nilsoberg:


1ndy (commented):
I tend to think about the scripts we write as user-facing, not backend tools for us to orchestrate analysis. This is because I think it gives the tools the best chance of persisting without a graphical front end. For that reason I try to name arguments in a way represents their purpose in the pipeline, not their purpose in connecting to a frontend. If I were a user trying to run these scripts, `--user-fasta-file` and `--user-accession-file` might be difficult options to understand on intuition alone.

nilsoberg:
The rationale for this issue/PR is for option clarity.  In new scripts I've been changing options to indicate if an option is a file or an input or output file, and I thought I could make this retroactive since the number of options will become more complicated over time as import features get added.  For example, in some scripts it may be beneficial to designate options as being a file (e.g. `--optionX-file`) or input and output if there are both (e.g. `--optionY-input-file` and `--optionY-output-file`).  

However, in retrospect I think this issue/PR shouldn't ever been opened and I'm inclined to reject the PR.  This could be a case where the Keep It Short and Simple principle might be applied.  `--help`, documentation, POD, or reading a script should be sufficient.  However, I am open to keeping the PR but changing the options to `--fasta-file` (instead of `--user-fasta-file`) -- assuming that was what you were meaning.

1ndy:
My comment was about continuing to develop the scripts/workflows as if they will be used by people, not just run as a backend to the site. For now I agree that writing good documentation is more important and this does not need to be merged.

PR #84, opened 2024-10-02
82 refactor color ssn utility for performance and code reuse

### LINKED ISSUE #82: Refactor Color SSN utility for performance and code reuse
Refactor the current Color SSN utility as follows:

- [x] Rewrite the XGMML parsing and writing code to reduce memory footprint
- [x] Refactor the Color SSN Nextflow pipeline so that the SSN coloring code can be shared with the GNT and Cluster Analysis pipelines
- [x] Refactor the `import` utilities in EST Nexflow pipeline so that they can be shared with the Color SSN, GNT, and Cluster Analysis pipelines

nilsoberg:


docs/source/pipelines/shared/python/compute_clusters.rst
```diff
@@ -0,0 +1,28 @@
+
+Compute Clusters
+================
+
+Clusters are computed based on node connectivity using igraph. The input is a simple
+edgelist file for performance reasons (i.e. the graph isn't contructed on the
```
1ndy: "contructed"

pipelines/colorssn/colorssn.nf
```diff
@@ -1,65 +1,8 @@
...
-    """
-}
+include { color_and_retrieve } from "../shared/nextflow/color_workflow.nf"
 
 workflow {
-    if (params.ssn_input =~ /\.zip/) {
-        ssn_file = unzip_input(params.ssn_input)
-    } else {
-        ssn_file = params.ssn_input
-    }
```
1ndy, line 62: Would it be possible to test for the magic number (`PK\x03\x04` I think) as well as the extension? I'm just thinking about a user misnaming their file, possibly maliciously

pipelines/shared/nextflow/color_workflow.nf
```diff
@@ -0,0 +1,202 @@
...
+        fasta_file=\${fasta_file/#$id_list_dir/fasta}
+        perl $projectDir/../shared/perl/get_sequences.pl --fasta-db ${params.fasta_db} --sequence-ids-file \${id_file} --output-sequence-file \${fasta_file}
+    done
+    for id_file in $id_list_dir/uniref50_ids/*.txt; do
+        [ -e "\${id_file}" ] || continue
+        fasta_file=\${id_file/%.txt/.fasta}
+        fasta_file=\${fasta_file/#$id_list_dir/fasta}
+        perl $projectDir/../shared/perl/get_sequences.pl --fasta-db ${params.fasta_db} --sequence-ids-file \${id_file} --output-sequence-file \${fasta_file}
+    done
+    """
```
1ndy: I would try to make `get_fasta` accept a tuple of the path to the idlist and the verison tag. In the `workflow` section, you could then filter for files that exist and aggregate by version tag and pass all of the id files for that version to the process or pass every pair of (file, version) to `get_fasta` to parallelize all of the lookups. Either way I would try to avoid the explicit for loops and manual directory management. I can elaborate on this if needed.

pipelines/shared/nextflow/color_workflow.nf
```diff
@@ -0,0 +1,202 @@
...
+        path "ssn____local.xgmml"
+    """
+    perl $projectDir/../shared/perl/unzip_xgmml_file.pl --in $ssn_zipped --out ssn____local.xgmml
+    """
+}
+
+// This is necessary to avoid Docker mounting the same directory twice; this occurs
+// because we need to mount the root of the project so that we can get access to the
+// lib directory. When using a SSN that is inside the project directory then the project
+// directory is also mounted.
```
1ndy: Is this mainly an issue in testing? If so could the tests be restructured to avoid the added complexity

pipelines/shared/nextflow/color_workflow.nf
```diff
@@ -0,0 +1,202 @@
...
+    output:
+        path 'id_lists/', emit: 'id_lists'
+        path 'cluster_sizes.txt', emit: 'cluster_sizes'
+    """
+    mkdir id_lists
+    mkdir id_lists/uniprot_ids id_lists/uniref90_ids id_lists/uniref50_ids
+    perl $projectDir/../shared/perl/get_id_lists.pl --cluster-map $cluster_id_map --singletons $singletons \
+        --uniprot id_lists/uniprot_ids --uniref90 id_lists/uniref90_ids --uniref50 id_lists/uniref50_ids \
+        --seqid-source-map $seqid_source_map --cluster-sizes cluster_sizes.txt \
+        --config ${params.efi_config} --db-name ${params.efi_db}
```
1ndy: Scripts should still be runnable without Nextflow, could you make Perl create these directories instead of doing it in the process? You could also output tuples instead of directory names so that each file is tagged with its uniref version.

pipelines/shared/nextflow/color_workflow.nf
```diff
@@ -0,0 +1,202 @@
...
+    publishDir params.final_output_dir, mode: 'copy'
+    input:
+        path cluster_id_map
+        path seqid_source_map
+        path singletons
+    output:
+        path 'stats.txt', emit: 'stats'
+    """
+    perl $projectDir/../shared/perl/compute_stats.pl --cluster-map $cluster_id_map --seqid-source-map $seqid_source_map \
+        --singletons $singletons --stats stats.txt
```
1ndy: `conv_ratio` and `compute_stats` both exist as python functions. I think they both accept parquet files so they obviously wouldn't work for this but in the future it would be good to define those computations in terms of nodes and edges exactly once and then only write code to adapt the input type to work with those functions. The calculations could be Perl or Python, the important thing would be the ability to read diverse file types.

pipelines/shared/nextflow/color_workflow.nf
1ndy: https://www.nextflow.io/docs/latest/channel.html#operators

pipelines/shared/python/compute_clusters.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def create_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(description="Compute clusters from a SSN edgelist")
+    add_args(parser)
+    return parser
+
+
+def get_node_id_mapping(id_map_file: str) -> tuple[dict, dict]:
+
```
1ndy, line 67: Add a docstring explaining the file format and what the mappings represent

pipelines/shared/python/compute_clusters.py
```diff
@@ -0,0 +1,82 @@
...
+    idx_node_size = {}
+
+    with open(id_map_file) as fh:
+        header = fh.readline()
+        for line in fh:
+            parts = line.rstrip().split("\t")
+            if len(parts) >= 2:
+                idx = int(parts[0])
+                idx_seqid_map[idx] = parts[1]
+                idx_node_size[idx] = int(parts[2]) if len(parts) >= 3 and parts[2] else 1
```
1ndy: stylistically I would avoid these kind of one-liner if statements, I don't see a lot of python code that uses them. Also, add a comparison for `parts[2]` to make it clear what you expect it to be

pipelines/shared/python/network.py
```diff
@@ -0,0 +1,181 @@
...
+        with open(info_file, "w") as fh:
+            fh.write("cluster_num_by_seq\tcluster_size_by_seq\tcluster_num_by_node\tcluster_size_by_node\n")
+            # The dict idx here represents the old (self.components) cluster number; sorting
+            # in the file will be done by sequence number
+            for idx in self.cluster_num_map_by_seq:
+                cluster_num_seq = self.cluster_num_map_by_seq[idx]
+                cluster_num_node = self.cluster_num_map_by_node[idx]
+                cluster_size_seq = self.cluster_size_by_seq[idx]
+                cluster_size_node = self.cluster_size_by_node[idx]
+                fh.write(f"{cluster_num_seq}\t{cluster_size_seq}\t{cluster_num_node}\t{cluster_size_node}\n")
```
1ndy, line 143: I don't have a strong suggestion on this but since everything is going to end up in a TSV, is there a way to construct the data as a table rather than in dictionaries? I might have missed something but for the most part it looks like you aren't doing random insertions/lookups so constructing parallel arrays and inserting into a pandas dataframe or arrow table might retain more of the semantic meaning of the data as its being processed. Just a thought.

pipelines/shared/nextflow/color_workflow.nf
nilsoberg: I didn't think about it, but yes, this is only an issue when the test data is inside the source code directory tree.  I didn't like my change and I feel it's a hack, so I'll remove this copy stuff and figure out something else for the tests.  The latter change will come in a separate issue/PR.

pipelines/shared/python/network.py
nilsoberg: I'm not exactly happy with the code since it's more of a brute force approach (although I don't consider it a hack).  Because the same clusters are ordered differently depending on metanode+cluster size or just cluster size I am storing the values in separate dictionaries.  I agree it could be different, but I need to move on to other things.  If I have to refactor this code in the future I'll take your suggestions into account.

pipelines/shared/python/compute_clusters.py
```diff
@@ -47,6 +47,22 @@ def create_parser() -> argparse.ArgumentParser:
...
+Parameters
+----------
+    id_map_file
+        path to the file that contains the mapping
+
+Returns
+-------
+    dict mapping node index to sequence ID
+    dict mapping node index to metanode size
+"""
```
1ndy: this should go underneath the function signature

1ndy (approved):

PR #86, opened 2024-10-04
Organize tests to support alternative database

### LINKED ISSUE #85: Support MySQL database basic tests
The `tests/runtests.sh` script doesn't currently support anything other than a simple dataset based on SQLite.  In order to better support tool development moving forward, tests need to be written that work with either a simple dataset or a larger EFI dataset and database based on MySQL.

nilsoberg:
Update logic for small test dataset retrieval
Parameters for database are set as environment variables for use by scripts in modules/


1ndy (approved):
lgtm

PR #95, opened 2024-10-28
64 relocate nf work log outputs in test

### LINKED ISSUE #64: Relocate nextflow work and log outputs in test
Each pipeline in the repo has a test in `tests/modules`. A module is simply a bash script which generate parameters and executes nextflow. However, the current `nextlfow run...` command does not specify where to store work and logs, and so the default `work/` and `.nextflow.log` are used. This pollutes the repository root and makes it difficult to investigate a failing test.

For each module in `tests/modules`:
- [x] Use the `-log` nextflow option to output the log file in the directory where results are stored for the pipeline
- [x] Use the `-work-dir` option on the `run` subcommand to ensure that the `work` directory is co-located with the results

More information on Nextflow CLI options can be found [here](https://www.nextflow.io/docs/latest/cli.html).

rbdavid:
- Relocated testing output to `/tmp/nextflow/efi_tests/` directory tree. 
- Used the `mktemp` command to create a unique directory tree for each module's tests; this setup allows for a user to run multiple rounds of tests without having to clean up the test directory trees via a `rm -rf $OUTPUT_DIR` call within each module's .sh file. Removed all calls to an `$OUTPUT_DIR` variable. The `mktemp` unique directory path is stored in the `$tmp_dir` var in `runtests.sh` and is passed as an argument to each module's .sh file.
- In `runtests.sh`, added a check for whether the `bash "tests/modules/$file" $tmp_dir $CONFIG_FILE` call was successful. If it was, the code will report that to the user and then clean up the test directory associated with that call. **Note:** the `rm -rf $tmp_dir` line is currently commented out so that the directory tree changes can be observed when tests are run. Upon accepting this change, I'll uncomment out this line so that cleaning of files is done automatically.   
- Added a `make tests-clean` recipe to the Makefile.

edited for clarity.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 all: docs build-pyEFI build-docker
 
-clean: docs-clean
+clean: docs-clean tests-clean
 
 build-pyEFI:
 	python -m build lib/pyEFI
@@ -35,4 +35,9 @@ test-pipelines:
 
 test-pyefi:
 	pytest lib/pyEFI
+
+tests-clean:
+	rm -rf /tmp/nextflow/efi_tests/
+	rm -rf .nextflow/
+	rm -rf tests/test_data/
 
--- a/Makefile
+++ b/Makefile
@@ -38,4 +38,6 @@ test-pyefi:
 
 tests-clean:
 	rm -rf .nextflow/
+	rm -rf tests/test_results/
+	rm -rf tests/test_data/
 
--- a/Makefile
+++ b/Makefile
@@ -39,5 +39,4 @@ test-pyefi:
 tests-clean:
 	rm -rf .nextflow/
 	rm -rf tests/test_results/
-	rm -rf tests/test_data/
 
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,5 @@ test-pyefi:
 	pytest lib/pyEFI
 
 tests-clean:
-	rm -rf /tmp/nextflow/efi_tests/
 	rm -rf .nextflow/
 
--- a/Makefile
+++ b/Makefile
@@ -39,5 +39,4 @@ test-pyefi:
 tests-clean:
 	rm -rf /tmp/nextflow/efi_tests/
 	rm -rf .nextflow/
-	rm -rf tests/test_data/
 
--- a/tests/runtests.sh
+++ b/tests/runtests.sh
@@ -23,10 +23,16 @@ fi
 
 echo "Using $CONFIG_FILE config files for processes"
 
-TEST_RESULTS_DIR=test_results
+# if $TEST_RESULTS_DIR not defined already then set to default path
+if [[ -z ${TEST_RESULTS_DIR} ]]; 
+then
+    TEST_RESULTS_DIR=tests/test_results
+fi
+
 if [ ! -d $TEST_RESULTS_DIR ]; then 
     mkdir -p $TEST_RESULTS_DIR
 fi
+echo "Test results will be written in $TEST_RESULTS_DIR"
 
 set +e
 
@@ -49,7 +55,7 @@ fi
 for file in $(ls tests/modules|grep '\.sh$'); do
     echo "================================================================================"
     echo "Executing tests in '$file'"
-    bash "tests/modules/$file" $TEST_RESULTS_DIR $CONFIG_FILE 2> >(tee $TEST_RESULTS_DIR/err.log >&2)
+    bash "tests/modules/$file" $TEST_RESULTS_DIR $CONFIG_FILE
     if [[ $? -eq 0 ]]; then
 	echo "Tests in '$file' passed"
     fi
--- a/tests/runtests.sh
+++ b/tests/runtests.sh
@@ -50,7 +50,7 @@ fi
 
 #bash "tests/modules/05_colorssn_uniprot.sh" $TEST_RESULTS_DIR $CONFIG_FILE
 #exit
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -36,3 +36,8 @@ test-pipelines:
 test-pyefi:
 	pytest lib/pyEFI
 
+tests-clean:
+	rm -rf /tmp/nextflow/efi_tests/
+	rm -rf .nextflow/
```
nilsoberg, line 40: Keep directory

Makefile
```diff
@@ -36,3 +36,8 @@ test-pipelines:
...
+	rm -rf tests/test_data/
```
nilsoberg: Keep test data `test_data/` around, but add `tests/test_data/` to `.gitignore` since it's part of the project (as opposed to a dev environment configuration)

nilsoberg (changes requested):
Revert changes to the module scripts, and add the `-log $OUTPUT_DIR/nextflow.log` and `-w $OUTPUT_DIR/work` arguments to the `nextflow` commands like in the commits in the PR.

Additionally, leave the `.nextflow/` directory alone as mentioned in the issue.  Don't remove it (e.g. strike it from the `tests-clean` make section).  

Add `test_results/` and `.nextflow/` to `.gitignore`.

Makefile
nilsoberg: Also add `test_results/` to `.gitignore`.

Makefile
rbdavid: As you suggested in slack, I could make multiple cleaning recipes in the Makefile. W/ the `mktemp` process that I developed, a `make tests-clean` call would only need to  happen very rarely if at all. So adding more cleaning recipes to the Makefile doesn't seem necessary.

rbdavid: From discussion https://github.com/EnzymeFunctionInitiative/EST/issues/64#issuecomment-2442120159. 
> 4. running a `make` command that works on an external hard-coded directory seems like a bad practice

I will update the `tests-clean` recipe to only remove repo-local files.

rbdavid:
So, my idea of the `tests-clean` recipe in the Makefile is a make call that returns the `tests/` subdir back to "pristine" condition. It should be called rarely. I think it should be removing `tests/test_data/` and `tests/test_results`. I also think at least one cleaning recipe should be removing the `.nextflow/` dir, which (AFAIK) is only produced inside the git repo directory when the tests are run. Maybe I revert the changes I made to the Makefile back to original state and create issue(s) centered around updating the Makefile? @nilsoberg any suggestions are appreciated here.

rbdavid:
In regards to commit [4369729](https://github.com/EnzymeFunctionInitiative/EST/pull/95/commits/436972940a70e3f37271ba3229a6c999045993b8), this enables a user to define TEST_RESULTS_DIR as an env variable to a path where the results of the tests will be written. This can be really helpful when we are running tests, checking for veracity of test results between branches, local changes, etc. An example of its use:

`source tests/test_env.sh && export TEST_RESULTS_DIR=/tmp/nextflow/efi_tests/issue64/ && bash tests/runtests.sh test.docker.config`

This will run the tests with all output (log files, results files, and work directories) written to `/tmp/nextflow/efi_tests/issue64/'. 

If the TEST_RESULTS_DIR variable is not declared before running `tests/runtests.sh`, then all those results files will be written in `tests/test_results/`.

tests/runtests.sh
```diff
@@ -46,8 +52,12 @@ fi
 
 #bash "tests/modules/05_colorssn_uniprot.sh" $TEST_RESULTS_DIR $CONFIG_FILE
```
nilsoberg, line 51: @nilsoberg This worked it's way into the main branch on mistake.  These two lines should be removed.

tests/runtests.sh
```diff
@@ -25,16 +23,24 @@ fi
 
 echo "Using $CONFIG_FILE config files for processes"
 
-rm -rf $TEST_RESULTS_DIR
-mkdir -p $TEST_RESULTS_DIR
+# if $TEST_RESULTS_DIR not defined already then set to default path
+if [[ -z ${TEST_RESULTS_DIR} ]];
```
nilsoberg: Double check, but I think the correct way to check if a variable is defined is with `-z ${TEST_RESULTS_DIR+1}` (see https://stackoverflow.com/a/13864829).

tests/runtests.sh
nilsoberg, line 1: The script uses both double and single brackets in if statements.  If possible, those should be consistent (I believe we use double brackets throughout the project - if not, that should be put into a "cleanup" issue so that the project consistent).

Makefile
nilsoberg: I think we should keep `.nextflow/` around and put it in the `.gitignore` file (as was already done).  Since test results are stored in `$TEST_RESULTS_DIR` (defaulting to `test_results/`), I would do something like this in the recipe:
`
if [[ -z ${TEST_RESULTS_DIR+1} && -d "${TEST_RESULTS_DIR}"]]; then
    rm -rf $TEST_RESULTS_DIR/*/*.log
    rm -rf $TEST_RESULTS_DIR/*/*_work
fi
`
Or
`
if [[ -z ${TEST_RESULTS_DIR+1} && -d "${TEST_RESULTS_DIR}"]]; then
    rm -rf $TEST_RESULTS_DIR
fi
`
@1ndy do you have an opinion on this?

nilsoberg (approved):

PR #119, opened 2025-01-13
116 ena build pipeline

### LINKED ISSUE #116: ENA build pipeline
Use Nextflow to develop a pipeline for building the ENA EFI database table from raw ENA datasets.  The pipeline should:

- [x] Unzip files and handle errors without having to restart the entire pipeline
- [x] Process files and handle errors without having to restart the entire pipeline
- [x] Merge output files
- [ ] Document cleanup process (but automate it)

nilsoberg:


bin/create_buildena_nextflow_params.py
```diff
@@ -0,0 +1,74 @@
...
+import os
+
+def add_args(parser: argparse.ArgumentParser):
+    """
+    add argumdents for GNT pipeline to ``parser``
+    """
+    parser.add_argument("--source-dir", type=str, required=True, help="Path to the root ENA dataset directory")
+    parser.add_argument("--output-dir", required=True, type=str, help="Location for results. Will be created if it does not exist")
+    parser.add_argument("--efi-config", required=True, type=str, help="Location of the EFI config file")
+    parser.add_argument("--efi-db", required=True, type=str, help="Name of the MySQL database to use (e.g. efi_202406) or name of the SQLite file")
```
1ndy, line 15: Keep the order of the kwargs consistent among all of these

docs/source/pipelines/buildena/index.rst
```diff
@@ -0,0 +1,37 @@
...
+
+The pipeline may then be executed with the file generated by this command::
+   
+   nextflow -C conf/buildena/slurm.config run pipelines/colorssn/colorssn.nf -params-file results/params.yml -w WORK_DIR
+
+It is vital that the ``WORK_DIR`` path be set to a location with a large storage
+capacity since many intermediate files are created, some of which can be large.
+
+Since the datasets are very large, use the ``-resume`` parameter to the
+``nextflow`` execution command.
```
1ndy, line 27: Can you add some more context for when to use resume? Are there circumstances under which a failed job needs to be restarted from the beginning

lib/EFI/ENA/IdMappingFile.pm
```diff
@@ -0,0 +1,100 @@
...
+    if (exists $self->{forward_map}->{$uniprotId}->{$idType}) {
+        return @{ $self->{forward_map}->{$uniprotId}->{$idType} };
+    } else {
+        return ();
+    }
+}
+
+
+sub finish {
+}
```
1ndy, line 97: Is this a placeholder for future work?

docs/source/pipelines/buildena/index.rst
```diff
@@ -0,0 +1,37 @@
+Build ENA
+=========
+
+The ENA database table for the EFI database is created using the European
+Nucleotide Archive dataset.
+
```
1ndy, line 6: Consider adding an example DAG image to this page

1ndy (commented):
There isn't documentation for the Perl modules - I'm okay with this if it will be written in a future issue.

There isn't a a script to generate a slurm file for this pipeline, is it meant to only be triggered manually? In the long term it might be good to bring it in line with other modules so that is can be scheduled to run the same way but I don't have strong feelings about it right now and am just curious.

lib/EFI/ENA/IdMappingFile.pm
nilsoberg: This module was copied from an old project and this is a leftover that I didn't remove when cleaning it up.

docs/source/pipelines/buildena/index.rst
nilsoberg: I don't know how helpful resuming is actually going to be.  Time will tell as the pipeline gets re-run for each release because it's hard to test at scale; right now I estimate the pipeline taking approximately a week to complete.  There are about 1.9m files in the input dataset as of 1/13/2025 and it took Nextflow about 1.5 days to find all of the files via `Channel.fromPath`.  The goal is that if the process is killed for whatever reason that resume will start the pipeline without having to identify all of the files again or uncompress files that are already uncompressed.  There are no obvious circumstances within a specific database build when it is desirable to start from the beginning (although each new database will need to start cleanly).

nilsoberg:
The Perl code will get refactored as part of a longer-term EFI database build pipeline rewrite, which is why this code is not as fleshed out as other parts and the module was not documented.

nilsoberg:
There is no Slurm script that is created like those created `create_nextflow_job.py`.  In the interim between now and when the EFI database build pipeline is written, the user is expected to trigger and monitor this from the command line, which would use Slurm internally to run the individual unzip and process jobs.  For now this is an in-house temporary solution.

I am fine if we keep the branch but put this PR on hold until the longer-term database build pipeline rewrite is complete (Q4 2025).

docs/source/pipelines/buildena/index.rst
1ndy: Ah, that makes sense. Not sure if speed is a concern right now but it is interesting that the file discovery taking so long - it looks like everything is a queue channel and so I would expect the unzipping to start almost immediately. If not, discovery of a higher level directory could be done more quickly in the workflow block (e.g. `wgs/public/**`) and individual files for each subdir could be found in parallel using Unix tools in a process block.

1ndy:
I don't have an issue with merging this now, remaining work fits neatly into a new issue so it won't be difficult to schedule improvements later on.

rbdavid (approved):
Next time this pipeline is run, we should add `-with-timeline` argument to the `nextflow run` call to get a sense of how long each step takes. This might require us to move the `Channel.fromPath` lines to a process(es) so that we can consider those steps as well.

rbdavid (changes requested):
Running this workflow on EFI cluster, we're seeing some flaws that only arise due to the scale of the ENA database. Some re-thinking needs to be done.

PR #121, opened 2025-01-23
114 add gnd workflow

### LINKED ISSUE #114: Add GND workflow
The GNT workflow has code that creates GNDs.  This code needs to be exposed via a new Nextflow pipeline in a way that doesn't depend on all of the other parts of the GNT workflow that are not relevant to GNDs.  It will involve refactoring one of the processes in the `pipelines/shared/nextflow/color_workflow.nf` script to a shared Nextflow script that can be used by the Color SSN workflow, GNT, and GND pipelines.

nilsoberg:
Added workflow for GND.  Due to documentation build, one of the rst files for GNT was also updated.


docs/source/lib/EFI/GNT/GND.pm.rst
```diff
@@ -0,0 +1,221 @@
...
+        // Start codon, but relative to the start of this sequence; for values
+        // in this table this will always be zero
+        rel_start integer
+        // End codon, but relative to the start of this sequence; for values
+        // in this table this will always be the sequence length
+        rel_stop integer
+        // Direction of the sequence, either 'normal' or 'complement'
+        direction varchar(10)
+        // Type of the sequence, either 'linear' or 'circular'
+        type varchar(8)
```
1ndy, line 144: Not exactly relevant to this PR but is there a check constraint in the actual table to enforce this? Since "type" is an ambiguous name this might add semantic meaning to the field. An enum type would be good too but I don't think SQLite supports those.

docs/source/lib/EFI/GNT/GND.pm.rst
```diff
@@ -0,0 +1,221 @@
...
+        anno_status integer
+        // Sequence description if SwissProt
+        desc text
+        // Pfam family description(s)
+        family_desc text
+        // InterPro family description(s)
+        ipro_family_desc text
+        // Sequence color, based on Pfam
+        color varchar(255)
+        // Many-to-one: many neighbors can have the same gene_key, all pointing to one
```
1ndy, line 218: At most one or exactly one?

docs/source/pipelines/gnd/gnd_database_schema.png
1ndy, line 1: I don't recognize the notation here as strictly ER or strictly UML but I think it gets the point across for now.

1ndy (commented):
There seems to be some inconsistency with `nextflow.preview.output` in the config files. Should it be present in all of the config files?

docs/source/lib/EFI/GNT/GND.pm.rst
nilsoberg: Exactly one.

bin/create_gnd_nextflow_params.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
+
+import argparse
+import glob
+import json
+import os
+
+def add_args(parser: argparse.ArgumentParser):
+    """
+    add argumdents for GND pipeline to ``parser``
```
rbdavid: "argumdents"

bin/create_gnd_nextflow_params.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    parser.add_argument("--output-dir", required=True, type=str, help="Location for results.  Must be empty, and will be created if it doesn't exist")
+    parser.add_argument("--cluster-id-map", required=True, type=str, help="The mapping of cluster numbers to IDs in the cluster for the GNDs")
+    parser.add_argument("--efi-config", required=True, type=str, help="Location of the EFI config file")
+    parser.add_argument("--efi-db", required=True, type=str, help="Name of the MySQL database to use (e.g. efi_202406) or name of the SQLite file")
+    parser.add_argument("--job-id", default=131, help="ID used when running on the EFI website. Not important otherwise")
```
rbdavid, line 16: "default=131", why is this the default? Why not use a uuid string or some other unique identifier? Just curious because this is also used in EFIToolsKBase code.

bin/create_gnd_nextflow_params.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def check_args(args: argparse.Namespace) -> argparse.Namespace:
+    """
+    Test file paths and rewrite them to be absolute
+    """
+    fail = False
+
+    if not os.path.exists(args.cluster_id_map):
```
rbdavid, line 25: Nothing explicitly wrong with your implementation. `os.path.isfile()` would more accurately test for the presence of the cluster_id_map file. If the args.cluster_id_map path pointed to a directory, the `.exists()` method would return True when that is uninteresting or a bug. Very rare to happen though.

bin/create_gnd_nextflow_params.py
```diff
@@ -0,0 +1,77 @@
...
+        print(f"SSN Input file '{args.cluster_id_map}' does not exist")
+        fail = True
+    
+    if not os.path.exists(args.efi_config):
```
rbdavid, line 29: As above.

bin/create_gnd_nextflow_params.py
```diff
@@ -0,0 +1,77 @@
...
+        print(f"EFI config file '{args.efi_config}' does not exist")
+        fail = True
+    
+    if os.path.exists(args.output_dir):
```
rbdavid, line 33: Could use `os.path.isdir()`

bin/create_gnd_nextflow_params.py
```diff
@@ -0,0 +1,77 @@
...
+        if len(os.listdir(args.output_dir)) > 0:
+            print(f"Output directory '{args.output_dir}' is not empty, refusing to create params.yml")
+            fail = True
+    else:
+        try:
+            os.makedirs(args.output_dir)
+        except Exception as e:
+            print(f"Could not create output directory '{args.output_dir}': {e}")
+            fail = True
+
```
rbdavid, line 43: extra white space

bin/create_gnd_nextflow_params.py
```diff
@@ -0,0 +1,77 @@
...
+
+    if fail:
+        print("Failed to render params template")
+        exit(1)
+    else:
+        args.output_dir = os.path.abspath(args.output_dir)
+        args.cluster_id_map = os.path.abspath(args.cluster_id_map)
+        args.efi_config = os.path.abspath(args.efi_config)
+        if os.path.exists(args.efi_db):
```
rbdavid, line 52: as above.

rbdavid: Also, what happens if this check fails? `args.efi_db` will remain the same value as input by the user. But that file doesn't exist, so passing a nonexistent local path to the efi_db file would cause a crash on the nextflow side of things, right?

bin/create_gnd_nextflow_params.py
```diff
@@ -0,0 +1,77 @@
...
+        "final_output_dir": output_dir,
+        "cluster_id_map": cluster_id_map,
+        "efi_config": efi_config,
+        "efi_db": efi_db
+    }
+    params_file = os.path.join(output_dir, "params.yml")
+    with open(params_file, "w") as f:
+        json.dump(params, f, indent=4)
+    print(f"Wrote params to '{params_file}'")
+    return params_file
```
rbdavid, line 72: No need for the return line since the returned object is not being used. But your implementation is fine too.

bin/create_nextflow_job.py
```diff
@@ -97,6 +105,8 @@ def create_parser() -> argparse.ArgumentParser:
         params_output = create_generatessn_nextflow_params.render_params(**args_dict)
     elif args.pipeline == "gnt":
         params_output = create_gnt_nextflow_params.render_params(**args_dict)
+    elif args.pipeline == "gnd":
```
rbdavid, line 108: These elif blocks look to be ordered alphabetically. So "gnd" should go before the "gnt" block. I'm also fine with organizing by flow of the EFI Tools. Or not reorganizing at all.

docs/source/lib/EFI/GNT/GND.pm.rst
```diff
@@ -0,0 +1,221 @@
...
+
+
+Parameters
+^^^^^^^^^^
+
+``$gnn``
+   A reference to a **EFI::GNT::GNN** object; the GNN data in ``$gnn``
+   should have already been retrieved.
+
+``$gndFile``
```
rbdavid: different parameter name than used in lines 61 and 98. On purpose?

docs/source/lib/EFI/GNT/GND.pm.rst
```diff
@@ -0,0 +1,221 @@
...
+   The path to a GND file to create.
+
+
+
+Returns
+^^^^^^^
+
+Returns 0 if there was an error or the file exists; 1 otherwise.
```
rbdavid, line 89: This might show my lack of understanding perl, but isn't a zero return code mean successful completion? Non-zero return codes signify failures. Can we match that expected behavior?

docs/source/pipelines/gnd/index.rst
```diff
@@ -0,0 +1,76 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The GND pipeline is configured through a YAML (or JSON) file containing
+parameters. The parameter file may be created manually or by running
+``bin/create_gnd_nextflow_params.py``. An example usage of the command: ::
+
+   python bin/create_gnd_nextflow_params.py --output-dir results/ --cluster-id-map id_list.txt --efi-config efi.config --efi-db efi_db.sqlite
+
+The pipeline uses the input ``id_list.txt`` file to extract genome
+context from the EFIcompute the GNNs. The other parameters are required
```
rbdavid: wording issue? What's this sentence trying to say?

docs/source/lib/EFI/GNT/GND.pm.rst
rbdavid: Also, if these exit codes have been used consistently throughout EFI code, then ignore my suggestion here. I haven't dug into how other EFI related perl codes exit.

bin/create_gnd_nextflow_params.py
1ndy: In production, each job receives a numeric ID to differentiate it. When this code is used to generate real jobs, that ID can be passed as an argument by the front end. Setting a default gave me one less thing to worry about in early testing. I chose 131 because it was my office number.

The ID shows up in graph titles and a few other places so John probably wouldn't like the look of a UUID.

bin/create_gnd_nextflow_params.py
1ndy: As noted in the help string, that parameter is not always a path. The tools work with both SQLite files and full MariaDB databases so sometimes it is a database name. Verifying that the name is correct would be a bit too expensive I believe so we don't currently do it.

bin/create_gnd_nextflow_params.py
1ndy: It's used in `create_nextflow_job.py`

bin/create_gnd_nextflow_params.py
1ndy: The convention in the entire project is `.exists()`, changing it is a valid conversation but probably not within this issue.

bin/create_gnd_nextflow_params.py
rbdavid: I'll raise this as an issue in EFIToolsKBase, whether the jobid can be left as an empty string or the narrative ID and version number (or something else completely). Currently, all graphs produced from the KBase EST apps have Jobid 131 in their titles. Just a bit of dissonance for how the EST pipeline is being run.

bin/create_gnd_nextflow_params.py
rbdavid: Just thought I should bring it to attention, since the "convention" is a potential source of bugs (albeit very rare to actually happen).

bin/create_gnd_nextflow_params.py
rbdavid: Good catch!

PR #128, opened 2025-02-03
Provide descriptive database connection error messages

nilsoberg:
Add database connection message into code.  Added database validation to scripts.  Moved database connection to highest level (e.g. at script instead of buried inside modules).


nilsoberg:
Closing since this PR was based on the wrong branch.

PR #129, opened 2025-02-04
Add nextflow.preview.output option to all config files

### LINKED ISSUE #122: Inconsistency with `nextflow.preview.output` in the config files
There seems to be some inconsistency with `nextflow.preview.output` in the config files. Should it be present in all of the config files?

_Originally posted by @1ndy in https://github.com/EnzymeFunctionInitiative/EST/pull/121#pullrequestreview-2571227767_

nilsoberg:
Made all `conf/*/*.config` files consistent by adding `nextflow.preview.output` to each file.

### DIFF
--- a/conf/colorssn/kbase.config
+++ b/conf/colorssn/kbase.config
@@ -1,6 +1,8 @@
 timeline.overwrite = true
 dag.overwrite = true
 report.overwrite = true
+nextflow.preview.output = true
+
 conda.enabled = true
 
 process {
--- a/conf/colorssn/pbspro.config
+++ b/conf/colorssn/pbspro.config
@@ -1,6 +1,7 @@
 timeline.overwrite = true
 dag.overwrite = true
 report.overwrite = true
+nextflow.preview.output = true
 
 executor {
     name = 'pbspro'
--- a/conf/colorssn/slurm.config
+++ b/conf/colorssn/slurm.config
@@ -1,6 +1,7 @@
 timeline.overwrite = true
 dag.overwrite = true
 report.overwrite = true
+nextflow.preview.output = true
 
 executor {
     name = 'slurm'
--- a/conf/est/docker.config
+++ b/conf/est/docker.config
@@ -1,6 +1,7 @@
 timeline.overwrite = true
 dag.overwrite = true
 report.overwrite = true
+nextflow.preview.output = true
 
 docker.enabled = true
 
--- a/conf/est/kbase.config
+++ b/conf/est/kbase.config
@@ -1,6 +1,8 @@
 timeline.overwrite = true
 dag.overwrite = true
 report.overwrite = true
+nextflow.preview.output = true
+
 conda.enabled = true
 
 process {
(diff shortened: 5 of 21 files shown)


rbdavid:
Just so it gets documented in the git history, there's been some breaking changes between Nextflow 24.04 and 24.10 in regards to "preview of workflow publishing" (https://nextflow.io/docs/latest/workflow.html#migrating-from-first-preview). 

Documentation for the 24.04's workflow publishing page is only accessible via their github branch (https://github.com/nextflow-io/nextflow/blob/v24.04.4/docs/workflow.md#publishing-outputs). Grumble grumble grumble. 

If I'm understanding their documentation correctly, 24.04 implemented the `nextflow.preview.output` config parameter, workflow/process `publish:` sections, and a workflow output block. As far as I can tell, we aren't using the `publish:` sections in any of the EST nextflow scripts nor the `output {` blocks. Instead, we're currently using `publishDir` lines inside of process blocks, for example: https://github.com/EnzymeFunctionInitiative/EST/blob/cdcc83fc5d52888281dc94f611e8089b60f56953/pipelines/gnt/gnt.nf#L5-L6 

Please correct me if I am wrong, but I don't think adding the `nextflow.preview.output` parameter is gonna change the behavior of our nextflow codes at the moment because we haven't ported our pipelines to using this new feature for 24.04 (now outdated as of 24.10).

rbdavid:
As far as I can tell, there's no harm in adding these parameters to the config files. We just need to be aware that the `nextflow.preview.output` parameter isn't immediately being used in the codes and that, if we were to utilize the feature, our nextflow codes may break or behave unexpectedly between nextflow versions 24.04 and 24.10+. Currently, our documentation for EST setup does not explicitly state that users should download 24.04 or 24.10. I'm running 24.04, but maybe I should be testing our workflows on multiple nextflow versions?

nilsoberg:
You're right, the `publishDir` feature is used in process blocks.  I've never liked the way the 24.04 "preview" `publishDir` feature works.  However, it serves a purpose for the pipelines that wasn't obvious when refactoring the color SSN, GNT, and GND pipelines.  Unfortunately, the 24.10 breaking changes will require a significant refactoring.  I don't see anyway within our time constraints to upgrade to 24.10 within the next several months so we're going to have to code for 24.04 for now.  We should keep the 24.10 updates in mind though as future features are developed so that refactoring doesn't have to happen for new pipelines.

nilsoberg:
> As far as I can tell, there's no harm in adding these parameters to the config files. We just need to be aware that the `nextflow.preview.output` parameter isn't immediately being used in the codes and that, if we were to utilize the feature, our nextflow codes may break or behave unexpectedly between nextflow versions 24.04 and 24.10+. Currently, our documentation for EST setup does not explicitly state that users should download 24.04 or 24.10. I'm running 24.04, but maybe I should be testing our workflows on multiple nextflow versions?

Just keep testing with 24.04.  It should go in the documentation somewhere.

rbdavid:
So, all that being said, PR looks good to me. Ready to be merged.

rbdavid (approved):
Tests pass. Looks good to me.

1ndy (approved):

PR #144, opened 2025-02-18
137 refactor est pipelines to share nextflow processes between pipelines

### LINKED ISSUE #137: Refactor EST pipelines to share Nextflow processes between pipelines
Several processes can be shared between pipelines (e.g. CGFP).  Refactor those processes out of EST into shared Nextflow scripts so that other pipelines can use those processes.

nilsoberg:
This PR changes the pipelines to share Nextflow processes between workflows by refactoring some processes out of the est pipeline into separate Nextflow files.  Additionally, singleton files are better supported, and better control of annotation output was added.

### DIFF
--- a/pipelines/identifycgfp/cdhit.py
+++ b/pipelines/identifycgfp/cdhit.py
@@ -1,0 +1,105 @@
+
+class CdHitParser:
+    """
+    Utility class for parsing CD-HIT files and returning a mapping of cluster ID
+    to IDs in the cluster.
+    """
+
+    def __init__(self, cdhit_cluster_file):
+        """
+        Create an object
+
+        Parameters
+        ----------
+            cdhit_cluster_file
+                path to a file containing CD-HIT cluster results (e.g. .clstr)
+        """
+        # Get CD-HIT cluster data
+        self.cdhit_clusters = self.parse_file(cdhit_cluster_file)
+
+    def parse_file(self, cdhit_file: str) -> dict:
+        """
+        Parses a CD-HIT file and returns a dictionary of CD-HIT clusters.
+        
+        An example of a CD-HIT file:
+    
+            >Cluster 268
+            0       146aa, >A0A6M4JL07... *
+            1       146aa, >A0A8E0SEA1... at 100.00%
+            2       146aa, >A0AAE2VGX6... at 100.00%
+            3       146aa, >A0AAP5KZA6... at 100.00%
+            4       146aa, >P45945... at 100.00%
+            >Cluster 270
+            0       146aa, >A0A9Q4HJ08... *
+            1       146aa, >A0AAP3CN64... at 100.00%
+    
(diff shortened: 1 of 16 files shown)


pipelines/shared/nextflow/sequence.nf
```diff
@@ -0,0 +1,34 @@
...
+    perl $projectDir/../shared/perl/get_sequences.pl --fasta-db ${fasta_db} --sequence-ids-file ${accession_ids} --output-sequence-file ${accession_ids}.fasta
+    """
+}
+
+process multiplex {
+    input:
+        path fasta_file
+    output:
+        path 'sequences.fasta', emit: 'fasta_file'
+        path 'sequences.fasta.clstr', emit: 'clusters'
```
rbdavid: Double quotes instead.

pipelines/shared/python/network.py
```diff
@@ -152,6 +152,7 @@ def save_singletons(self, singletons_file: str):
                 path to a file to output singleton list to
         """
         with open(singletons_file, "w") as fh:
+            fh.write("node_label\n")
```
rbdavid, line 155: Should the header line include a '#' character to denote it as a comment/header?

pipelines/identifycgfp/cdhit.py
```diff
@@ -0,0 +1,105 @@
...
+        cluster_id = ""
+        with open(cdhit_file, 'r') as f:
+            for line in f:
+                line = line.strip()
+                if line.startswith('>'):
+                    # Extract cluster ID
+                    cluster_id = line[1:]
+                    clusters[cluster_id] = []
+                else:
+                    m = re.search(r'^.+>(.+)\.\.\. (.*)$', line)
```
rbdavid: Maybe update the regex pattern if the 2nd group isn't going to be used? `r">(.+)\.\.\."` works for me for the example provided and is a bit simpler than the current pattern by ignoring any start of string pattern matching.

rbdavid (approved):
Looks good. Only minor comments.

pipelines/shared/python/network.py
nilsoberg: No, the pipelines assume the first line is a header and parse it without a header character.

pipelines/identifycgfp/cdhit.py
nilsoberg: I'll look into this, but I believe it's parsing for the case where there are one or more characters before the `>`.  You are right, the second group after the `\.\.\.` is not used so it should be removed.

rbdavid: Oh good point. I'm not well versed in the cd-hit results file format. There is commented-out code for parsing the 2nd group, so maybe this is still a useful pattern. Is there any instance where we'd be interested in gathering the sequence lengths as well as the ID and percent identity values?

Issue #111, opened 2025-01-06
Support UniRef sequences throughout pipelines


nilsoberg:
Implement support for UniRef sequences throughout the pipelines.
- [x] `est` pipeline - filtering of sequences
- [x] `gnt` pipeline - metadata, tables, statistics
- [x] `colorssn` pipeline - sequence and ID list retrieval
- [x] Update `EFI::SSN::XgmmlWriter::AttributeHandler::GNT` with to properly support list node attributes when using UniRef
- [x] Update `pipelines/gnt/create_gnns.pl` and `pipelines/gnd/create_gnd.pl`

PR #146, opened 2025-03-03
145 missing input parameters for the gnt.nf pipeline

### LINKED ISSUE #145: Missing input parameters for the gnt.nf pipeline.
The perl code that is called within the gnt.nf `create_gnns()` process accepts input arguments `--nb-size` and `--cooc-threshold` but these arguments are not included in the call of this perl code in the nextflow process. For me to fully implement the mapping of KBase UI features with the apps' backend code (that calls the gnt.nf script), these parameters need to be included in the nextflow code as possible input arguments. These parameters will also need to be included in the `create_gnt_nextflow_params.py` script so that the `parameters.yml` file contains all required input arguments.

rbdavid:
Implement the `nb_size` and `cooc_threshold` input parameters for both the gnt.nf and gnd.nf code. Also, specifically use their input arguments in the perl backend codes.

### DIFF
--- a/bin/create_gnd_nextflow_params.py
+++ b/bin/create_gnd_nextflow_params.py
@@ -50,12 +50,13 @@ def create_parser():
     add_args(parser)
     return parser
 
-def render_params(cluster_id_map, efi_config, efi_db, output_dir, job_id, nextflow_config=None):
+def render_params(cluster_id_map, efi_config, efi_db, nb_size, output_dir, job_id, nextflow_config=None):
     params = {
         "final_output_dir": output_dir,
         "cluster_id_map": cluster_id_map,
         "efi_config": efi_config,
-        "efi_db": efi_db
+        "efi_db": efi_db,
+        "nb_size": nb_size
     }
     params_file = os.path.join(output_dir, shared_args.PARAMS_NAME)
     with open(params_file, "w") as f:
--- a/bin/create_gnd_nextflow_params.py
+++ b/bin/create_gnd_nextflow_params.py
@@ -14,6 +14,7 @@ def add_args(parser: argparse.ArgumentParser):
     Add arguments for GND pipeline to ``parser``
     """
     parser.add_argument("--cluster-id-map", required=True, type=str, help="The mapping of cluster numbers to IDs in the cluster for the GNDs")
+    parser.add_argument("--nb-size", type=int, required=False, default=20, help="Optional number of neighbors on the left and right of the input IDs to include in the analysis, an integer > 0.")
     shared_args.add_args(parser)
 
 def check_args(args: argparse.Namespace) -> argparse.Namespace:
@@ -33,6 +34,10 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
         print(f"SSN Input file '{args.cluster_id_map}' does not exist")
         fail = True
     
+    if args.nb_size < 0 or args.nb_size > 20:
+        print(f"Neighborhood size (--nb-size) was given a bad value ({args.nb_size}).")
+        fail = True
+
     if fail:
         print("Failed to render params template")
         exit(1)
--- a/bin/create_gnd_nextflow_params.py
+++ b/bin/create_gnd_nextflow_params.py
@@ -34,8 +34,8 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
         print(f"SSN Input file '{args.cluster_id_map}' does not exist")
         fail = True
     
-    if args.nb_size < 0 or args.nb_size > 20:
-        print(f"Neighborhood size (--nb-size) was given a bad value ({args.nb_size}).")
+    if args.nb_size < 1 or args.nb_size > 20:
+        print(f"Invalid value for --nb-size ({args.nb_size}).")
         fail = True
 
     if fail:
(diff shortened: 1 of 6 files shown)


rbdavid:
Local tests of 09_gnt.sh and 13_gnd.sh are passing for both docker.config and singularity.config files.

bin/create_gnt_nextflow_params.py
```diff
@@ -51,13 +61,15 @@ def create_parser():
...
 
-def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, job_id, nextflow_config=None):
+def render_params(ssn_input, efi_config, efi_db, fasta_db, nb_size, cooc_threshold, output_dir, job_id, nextflow_config=None):
     params = {
         "final_output_dir": output_dir,
         "ssn_input": ssn_input,
         "efi_config": efi_config,
         "efi_db": efi_db,
-        "fasta_db": fasta_db
+        "fasta_db": fasta_db,
```
rbdavid, line 70: I'm not sure if this parameter is needed for this workflow. I didn't see it used in the perl code.

rbdavid: `fasta_db` is used in the `color_and_retrieve()` process.

rbdavid: So, this may be another instance where the option for the user to choose between the UniProt, UniRef90, or UniRef50 sequence databases. But, this isn't an active parameter that is selected on the website so we need code to detect which sequence database to use by looking at the cluster sequence annotations/headers?

bin/create_gnd_nextflow_params.py
```diff
@@ -33,6 +34,10 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
         print(f"SSN Input file '{args.cluster_id_map}' does not exist")
         fail = True
     
+    if args.nb_size < 0 or args.nb_size > 20:
```
nilsoberg: `nb_size` should be 1-20, so `if args.nb_size < 1`.  Also might say "Invalid value for --nb-size ({args.nb_size})".  I think that wording is in line with what I say in Perl (maybe double check to see what the wording is with other Python scripts first).

bin/create_gnt_nextflow_params.py
```diff
@@ -38,6 +40,14 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
         print(f"FASTA database '{args.fasta_db}' not found")
         fail = True
 
+    if args.nb_size < 0 or args.nb_size > 20:
```
nilsoberg: As above

bin/create_gnt_nextflow_params.py
nilsoberg: This is for another issue (#111).  I don't believe the sequence version is necessary here because the code will detect which sequences are used and then look up the proper identifiers based on the input sequence type, so this argument should be sufficient.

pipelines/gnt/create_gnns.pl
```diff
@@ -40,7 +40,7 @@
 my $idMap = parse_cluster_map_file($opts->{cluster_map});
 
 my $gntAnno = new EFI::GNT::Annotations(dbh => $dbh);
-my $gnn = new EFI::GNT::GNN(dbh => $dbh, seq_cluster_id_map => $idMap, gnt_anno => $gntAnno);
+my $gnn = new EFI::GNT::GNN(dbh => $dbh, seq_cluster_id_map => $idMap, gnt_anno => $gntAnno, neighborhood_size => $opts->{nb_size});
```
nilsoberg, line 43: Is cooccurrence threshold used anywhere as an input to the GNN and GND scripts?

nilsoberg (changes requested):

bin/create_gnd_nextflow_params.py
rbdavid: I don't think there are other instances where numerical input values are validated in the "create_[pipeline]_params.py" scripts. Maybe we should be doing those validation checks, but that's for a new issue. I like your wording. Changed.

pipelines/gnt/create_gnns.pl
rbdavid: Yes, for the gnn analysis, line 48 of this same script. 
```
my $hubs = new EFI::GNT::GNN::Hubs(gnn => $gnn, cooc_threshold => $opts->{cooc_threshold}, seq_cluster_id_map => $idMap);
```

As far as I know, the cooccurrence threshold param is not relevant to the gnd.nf script because that workflow is not running the analysis to create cluster and pfam GNNs.

bin/create_gnt_nextflow_params.py
rbdavid: Just to clarify, the fix for #111 will enable the code to detect whether sequences in an SSN originate from UniProt, UniRef90, or UniRef50 sequence databases, correct? 

So, the "fasta_db" input parameter just needs to map to the directory path where the blast database files are stored. If I recall correctly, the three databases (all the associated files) are stashed in the same subdirectory in our EFI_DB reference data structure, right? I guess I'm asking specifically for KBase purposes because right now this input parameter is mapped to the UniProt seq database.

nilsoberg: The `fasta_db` parameter maps to a BLAST database which is the first part of the path to the files that make up a BLAST database.  For example, if a database named `combined.fasta` may or may not have a file called `combined.fasta` in a directory, but will for sure contain files named `combined.fasta.phr`, `.pin`, `.psd`, `.psi`, and `.psq` (or `combined.fasta.00.phr`, `combined.fasta.00.pin` ... depending on the size of the database).  (BLAST can also use a FASTA file as the database, it's just not as fast.).  The code doesn't check for the actual existence of the database, just verifies that the parameter exists.

nilsoberg: #111 will allow the code to detect the sequence source, yes.

nilsoberg (approved):

PR #170, opened 2025-04-21
168 port download efi db files

### LINKED ISSUE #168: Port code for downloading the large EFI DB files.
Code to download the split EFI DB files is currently housed in the EFIToolsKBase repo. 
- [x] Move that to this repo.
- [x] Implement a simple bash script that wraps up the commands (see `scripts/entrypoints.sh` in EFIToolsKBase) into one simple script. 
- [x] Document the download process. Try to include any notes about downloads failing/being interrupted/etc.

rbdavid:
Ported code from the EFIToolsKBase repo that is used to download blastdb, diamonddb, and the sqlite-formatted EFI DB files. The `download_file.py` code handles the requests of files from the EFI downloads server and joining of split files (if present). The `scripts/download_latest_efi_db.sh` script calls this python script three times, once for each of the important databases that need to be gathered.   

2025-04-21, RBD: Labeled this PR as WIP to get input from Nils. 

To do: 
- [x] check that these two directories are the correct home for these scripts. 
- [x] add documentation on how to run the `scripts/download_latest_efi_db.sh` script. 
- [x] add documentation about the structure of the database files on the EFI downloads page

Actual implementation of this code may depend on the HPC machine being used. Ideally, this download process does not happen on the head/login node; sbatch'ing or qsub'ing the bash script just requires the necessary header lines to be added. This assumes that compute nodes have access to internet though.


scripts/download_latest_efi_db.sh
nilsoberg, line 1: Will this file be documented somewhere?  Is that documentation even needed?

nilsoberg (approved):

scripts/download_latest_efi_db.sh
rbdavid: I've more formally implemented the input arguments for the `downloads_efi_dbs.sh` script (also renamed the file). Documentation in `docs/source/guides/databases.rst` now includes information about the databases we host and how the script works to download those databases.

docs/source/guides/databases.rst
```diff
@@ -69,12 +72,59 @@ Sequence Database Set Installation
...
-different than the system that they are started up on.
+different than the system that they are started up on. See the section below
+on Database Downloading for further details on accessing these sequence 
+databases. 
+
+Database Downloading
+====================
+As mentioned above, the metadata and sequence databases are available for 
+download at 
+https://efi.igb.illinois.edu/downloads/databases/ where the lastest as well as
```
nilsoberg: "latest" not "lastest"

docs/source/guides/databases.rst
```diff
@@ -69,12 +72,59 @@ Sequence Database Set Installation
...
+a year month date format (YYYYMMDD). The ``VERSION.txt`` file in each directory
+contains the metadata associated with the provided database files, including 
+the EFI DB version name (e.g. IP104) as well as the UniProt and InterPro 
+release IDs. 
+
+The metadata database consists of a single gzip'd sqlite file that is roughly
+80 GB in size. The BLAST and DIAMOND database files are tar'd directories and 
+are both roughly 185 GB. Unzipping and untarring these files requires close to 
+1 TB of available storage. To avoid potential errors or corrupted data while 
+downloading these large database files, we also provide splits of the full
```
nilsoberg: Better to not use "we" and instead something like "splits of the full ... are provided", or "EFI-Illinois" (or something like that) instead of "we".  That fits more or less with existing conventions.

PR #172, opened 2025-04-22
Sequence BLAST - retrieval options differentiated from all-to-all options

### LINKED ISSUE #126: Add a second e-value cutoff parameter for EST Sequence BLAST
For the Sequence BLAST path for EFI-EST, there are two `blastall` calls.  One happens during the initial BLAST sequence search (`get_sequence_ids()` process in est.nf) and the other during the all-to-all BLAST alignment step (`all_by_all_blast` process in est.nf). For both process, the same `params.blast_evalue` parameter is currently used although the webserver has two separate parameter input spaces, one for each blastall, see attached screenshots.  

Image
and 
Image

Should we be included two e-value cutoff parameters in the `est.nf` parameters? I've only skimmed the old codebase. I don't see much to indicate whether the two values were used independently or one was synonymous with the other.

### LINKED ISSUE #127: Update default value for params.num_blast_matches in est.nf
In a similar but slightly different vein to #126 , should the `params.num_blast_matches` even be used in the `all_by_all_BLAST` process? If a sequence is highly similar to a large number of sequences but the `params.num_blast_matches` value is sufficiently small, then some similar sequences will be outright ignored. Wouldn't we want to know all alignment score values during the `all_by_all_BLAST` alignment calculations?

EDIT: Default `--blast-matches` parameter for the `create_est_nextflow_params.py` code is 250. That seems like a very bad choice for that default value.

rbdavid:
This will close the immediate coding concerns about issues #126 and #127. The open questions about appropriate default values for all of those issues still stand (and can be turned into "question" issues). 

- [x] Implemented command line argument parameters for the `bin/create_est_nextflow_params.py` script for the Option A (Sequence BLAST) import mode, following a naming scheme `import_blast_*`. 
- [x] Moved other import mode-specific arguments to this same naming scheme. 
- [x] Mirrored the necessary `params.import_blast_*` naming in the est.nf script.

### DIFF
--- a/bin/create_est_nextflow_params.py
+++ b/bin/create_est_nextflow_params.py
@@ -34,7 +34,7 @@ def add_args(parser: argparse.ArgumentParser):
     # option A: Sequence BLAST
     blast_parser = subparsers.add_parser("blast", help="Import sequences using the single sequence BLAST option", parents=[common_parser]).add_argument_group("Sequence BLAST Options")
     blast_parser.add_argument("--blast-query-file", required=True, type=str, help="The file containing a single sequence to use for the initial BLAST to obtain sequences")
-    blast_parser.add_argument("--import-blast-fasta-db", type=str, help="FASTA file or BLAST database to use for the initial import to find sequences; must be set if the --sequence-version is uniref50 or uniref90")
+    blast_parser.add_argument("--import-blast-fasta-db", type=str, help="FASTA file or BLAST database to use for the initial import to find sequences; must be set if the --sequence-version is uniref50 or uniref90; defaults to the same as --fasta-db.")
     blast_parser.add_argument("--import-blast-num-matches", default=1000, type=int, help="Maximum number of matches returned by BLAST when retrieving sequences")
     blast_parser.add_argument("--import-blast-evalue", default="1e-5", help="Cutoff e-value to use in the BLAST sequence alignment when retrieving sequences")
 
--- a/bin/create_est_nextflow_params.py
+++ b/bin/create_est_nextflow_params.py
@@ -21,7 +21,7 @@ def add_args(parser: argparse.ArgumentParser):
     common_parser.add_argument("--accession-shards", default=16, type=int, help="Number of files to split Accessions list into. File is split so that sequence retrieval can be parallelized")
     common_parser.add_argument("--fasta-db", type=str, required=True, help="FASTA file or BLAST database to retrieve sequences from")
     common_parser.add_argument("--multiplex", action="store_true", help="Use CD-HIT to reduce the number of sequences used in analysis")
-    common_parser.add_argument("--blast-num-matches", default=250, type=int, help="Number of matches BLAST should return; dev site only?")
+    common_parser.add_argument("--blast-num-matches", default=250, type=int, help="Number of matches BLAST should return")
     common_parser.add_argument("--blast-evalue", default="1e-5", help="Cutoff E value to use in all-by-all BLAST")
     common_parser.add_argument("--sequence-version", type=str, default="uniprot", choices=["uniprot", "uniref90", "uniref50"])
     common_parser.add_argument("--filter", action="append", type=str, help="Filter sequences, use multiple times to indicate filter types")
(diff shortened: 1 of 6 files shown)


bin/create_est_nextflow_params.py
```diff
@@ -19,9 +19,9 @@ def add_args(parser: argparse.ArgumentParser):
     common_parser.add_argument("--duckdb-threads", default=1, type=int, help="Number of threads DuckDB can use. More threads means higher memory usage")
     common_parser.add_argument("--fasta-shards", default=128, type=int, help="Number of files to split FASTA into. File is split so that BLAST can be parallelized")
     common_parser.add_argument("--accession-shards", default=16, type=int, help="Number of files to split Accessions list into. File is split so that sequence retrieval can be parallelized")
-    common_parser.add_argument("--blast-matches", default=250, type=int, help="Number of matches BLAST should return")
     common_parser.add_argument("--fasta-db", type=str, required=True, help="FASTA file or BLAST database to retrieve sequences from")
     common_parser.add_argument("--multiplex", action="store_true", help="Use CD-HIT to reduce the number of sequences used in analysis")
+    common_parser.add_argument("--blast-num-matches", default=250, type=int, help="Number of matches BLAST should return")
```
nilsoberg: Perhaps phrase this something like "Number of matches BLAST should return in the all-by-all computation"

bin/create_est_nextflow_params.py
```diff
@@ -33,8 +33,10 @@ def add_args(parser: argparse.ArgumentParser):
     
     # option A: Sequence BLAST
     blast_parser = subparsers.add_parser("blast", help="Import sequences using the single sequence BLAST option", parents=[common_parser]).add_argument_group("Sequence BLAST Options")
-    blast_parser.add_argument("--blast-query-file", required=True, type=str, help="The file containing a single sequence to use for the initial BLAST to obtain sequences")
+    blast_parser.add_argument("--import-blast-query-file", required=True, type=str, help="The file containing a single sequence to use for the initial BLAST to obtain sequences")
```
nilsoberg: I would keep this as `--blast-query-file` because it specifically refers to the input to the `get_sequences` process; we don't add `import` prefix to the other options.  The import prefix on `--import-blast-num-matches` and `--import-blast-evalue` is important to distinguish from the main computation.

bin/create_est_nextflow_params.py
```diff
@@ -33,8 +33,10 @@ def add_args(parser: argparse.ArgumentParser):
...
     blast_parser.add_argument("--import-blast-fasta-db", type=str, help="FASTA file or BLAST database to use for the initial import to find sequences; must be set if the --sequence-version is uniref50 or uniref90")
+    blast_parser.add_argument("--import-blast-num-matches", default=1000, type=int, help="Maximum number of matches BLAST should return for the input query sequence")
```
nilsoberg: Perhaps phrase this "Maximum number of matches returned by BLAST computation when retrieving sequences"

bin/create_est_nextflow_params.py
```diff
@@ -33,8 +33,10 @@ def add_args(parser: argparse.ArgumentParser):
...
+    blast_parser.add_argument("--import-blast-evalue", default="1e-5", help="Cutoff e-value to use in the input BLAST sequence alignment.")
```
nilsoberg: Perhaps phrase this something like "Cutoff e-value to use in BLAST sequence alignment when retrieving sequences"

nilsoberg (approved):
It looks good, other than I would keep `--blast-query-file` rather than `--import-blast-query-file`

tests/modules/02a_est_blast.sh
```diff
@@ -11,7 +11,7 @@ OUTPUT_DIR="$TEST_RESULTS_DIR/$self"
 
 rm -rf $OUTPUT_DIR
 
-./bin/create_est_nextflow_params.py blast --output-dir $OUTPUT_DIR --efi-config $EFI_CONFIG_FILE --fasta-db $EFI_FASTA_DB --efi-db $EFI_DB_NAME --blast-query-file $EFI_TEST_BLAST_SEQ --nextflow-config $NXF_EST_CONFIG_FILE
+./bin/create_est_nextflow_params.py blast --output-dir $OUTPUT_DIR --efi-config $EFI_CONFIG_FILE --fasta-db $EFI_FASTA_DB --efi-db $EFI_DB_NAME --blast-query-file $EFI_TEST_BLAST_SEQ --nextflow-config $NXF_EST_CONFIG_FILE --import-blast-num-matches 250 --import-blast-evalue 1e-5 --import-blast-fasta-db $EFI_DB_NAME
```
rbdavid: Explicitly calling the `--import-blast-num-matches` argument here is needed to match previous runs of this test since the default value for this parameter is now different from the default value of the `--num-matches` parameter. I added the other input arguments to demonstrate their use.

nilsoberg: We need an example of this that doesn't include those parameters since they're not required.  I look at the tests to see the syntax and I imagine others might do that as well.  We're not comparing against previous results, right?  If that's the case then adding these parameters here doesn't matter and the default is fine, right?

rbdavid: Agreed. I'll reinstate the old form of this test but copy this version to `extended/` and rename it.

nilsoberg (approved):

PR #173, opened 2025-04-23
169 support generic job script creation

### LINKED ISSUE #169: Support generic job script creation
In order to support the capability to run the tools on multiple platforms it is necessary to refactor the support of the current template system to support inclusion of environment variables.  For example, on an EFI cluster a series of `module load` directives are required in each job that is submitted to the cluster.  When this is run on a CLI or on another cluster this environment will be different.

Support generic job script creation by adding environment inclusion.

nilsoberg:
Updates all job script generation scripts in `bin/create*.py` to accept Jinja templates for generating shell scripts to run Nextflow.  The Jinja templates are designed to be copied and modified to include implementation-specific environment variables and configurations in the output `run_nextflow.sh` script.  Also requires presence of UniRef and RepNode SSNs for Color SSN and GNT tests.

### DIFF
--- a/bin/create_colorssn_nextflow_params.py
+++ b/bin/create_colorssn_nextflow_params.py
@@ -38,6 +38,9 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
         print(f"FASTA database '{args.fasta_db}' not found")
         fail = True
 
+    if args.workflow_def is None:
+        args.workflow_def = os.path.abspath(NXF_SCRIPT)
+
     if fail:
         print("Failed to render params template")
         exit(1)
@@ -51,8 +54,7 @@ def create_parser() -> argparse.ArgumentParser:
     add_args(parser)
     return parser
 
-def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, job_id,
-        nextflow_config=None, templates_dir=None, template=None):
+def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, **kwargs: dict):
     params = {
         "final_output_dir": output_dir,
         "ssn_input": ssn_input,
@@ -69,5 +71,5 @@ def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, job_id,
 if __name__ == "__main__":
     args = check_args(create_parser().parse_args())
     params_file = render_params(**vars(args))
-    shared_args.save_run_script(args, workflow_def=NXF_SCRIPT, params_file=params_file)
+    shared_args.save_run_script(args, workflow_def=args.workflow_def, params_file=params_file)
 
--- a/bin/create_colorssn_nextflow_params.py
+++ b/bin/create_colorssn_nextflow_params.py
@@ -52,7 +52,7 @@ def create_parser() -> argparse.ArgumentParser:
     return parser
 
 def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, job_id,
-        nextflow_config=None, templates_dir=None, template=None, template_env=None):
+        nextflow_config=None, templates_dir=None, template=None):
     params = {
         "final_output_dir": output_dir,
         "ssn_input": ssn_input,
--- a/bin/create_colorssn_nextflow_params.py
+++ b/bin/create_colorssn_nextflow_params.py
@@ -51,7 +51,8 @@ def create_parser() -> argparse.ArgumentParser:
     add_args(parser)
     return parser
 
-def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, job_id, nextflow_config=None):
+def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, job_id,
+        nextflow_config=None, templates_dir=None, template=None, template_env=None):
     params = {
         "final_output_dir": output_dir,
         "ssn_input": ssn_input,
@@ -67,6 +68,6 @@ def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, job_id, n
 
 if __name__ == "__main__":
     args = check_args(create_parser().parse_args())
-    render_params(**vars(args))
-    shared_args.save_run_script(args, NXF_SCRIPT)
+    params_file = render_params(**vars(args))
+    shared_args.save_run_script(args, workflow_def=NXF_SCRIPT, params_file=params_file)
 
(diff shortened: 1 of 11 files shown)


bin/create_colorssn_nextflow_params.py
```diff
@@ -51,7 +51,8 @@ def create_parser() -> argparse.ArgumentParser:
     add_args(parser)
     return parser
 
-def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, job_id, nextflow_config=None):
+def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, job_id,
+        nextflow_config=None, templates_dir=None, template=None):
```
rbdavid: The `templates_dir` and `template` arguments aren't used in this function. Instead of explicitly defining them and other unused input variables, you can use `**kwargs` as a catch-all for input argument. Ex: 
```
def render_params(ssn_input, efi_config, efi_db, fasta_db, output_dir, **kwargs: dict):
    # handle all of the explicitly defined arguments
    # now handle kwargs
    for key, value in kwargs.items():
        print(f"{key}:{value}") # or do something else, depending on the key...
```

Using kwargs enables the function interface that you've designed for, while only explicitly defining in the `def` line the input arguments that are explicitly needed for this version of the function.

rbdavid: And, `**kwargs` plays nicely with the call of this function `render_params(**vars(args))` where you are converting the `args` argparse.Namespace object to a dict, where keys and values are gathered from Namespace's attributes. `**vars(args)` unpacks that dict and any undefined input arguments in that dict will be gathered/contained in the render_params `**kwargs`.

bin/create_est_nextflow_params.py
```diff
@@ -123,8 +123,8 @@ def render_params(output_dir, duckdb_memory_limit, duckdb_threads, fasta_shards,
                   accessions_file=None,
                   blast_query_file=None,
                   import_blast_fasta_db=None,
-                  nextflow_config=None
-                  ):
+                  nextflow_config=None,
+                  templates_dir=None, template=None):
```
rbdavid: See my note about `**kwargs`.

bin/create_generatessn_nextflow_params.py
```diff
@@ -104,7 +104,10 @@ def create_parser():
     add_args(parser)
     return parser
 
-def render_params(blast_parquet, fasta_file, seq_meta_file, output_dir, filter_parameter, filter_min_val, min_length, max_length, ssn_name, ssn_title, maxfull, uniref_version, efi_config, db_version, job_id, efi_db, mode, est_output_dir=None, nextflow_config=None):
+def render_params(blast_parquet, fasta_file, seq_meta_file, output_dir, filter_parameter,
+        filter_min_val, min_length, max_length, ssn_name, ssn_title, maxfull, uniref_version,
+        efi_config, db_version, job_id, efi_db, mode,
+        est_output_dir=None, nextflow_config=None, templates_dir=None, template=None):
```
rbdavid: See my note about `**kwargs`.

bin/create_gnd_nextflow_params.py
```diff
@@ -50,7 +50,8 @@ def create_parser():
     add_args(parser)
     return parser
 
-def render_params(cluster_id_map, efi_config, efi_db, nb_size, output_dir, job_id, nextflow_config=None):
+def render_params(cluster_id_map, efi_config, efi_db, nb_size, output_dir, job_id,
+        nextflow_config=None, templates_dir=None, template=None):
```
rbdavid: See my note about `**kwargs`.

bin/create_gnt_nextflow_params.py
```diff
@@ -61,7 +61,8 @@ def create_parser():
     add_args(parser)
     return parser
 
-def render_params(ssn_input, efi_config, efi_db, fasta_db, nb_size, cooc_threshold, output_dir, job_id, nextflow_config=None):
+def render_params(ssn_input, efi_config, efi_db, fasta_db, nb_size, cooc_threshold, output_dir, job_id,
+        nextflow_config=None, templates_dir=None, template=None):
```
rbdavid: See my note about `**kwargs`.

bin/create_nextflow_job.py
```diff
@@ -41,88 +42,62 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
...
+        workflow_def = create_est_nextflow_params.NXF_SCRIPT
+    elif pipeline == "generatessn":
+        params_file = create_generatessn_nextflow_params.render_params(**vars(args))
+        workflow_def = create_generatessn_nextflow_params.NXF_SCRIPT
+    elif pipeline == "gnt":
+        params_file = create_gnt_nextflow_params.render_params(**vars(args))
+        workflow_def = create_gnt_nextflow_params.NXF_SCRIPT
+    elif pipeline == "gnd":
+        params_file = create_gnd_nextflow_params.render_params(**vars(args))
+        workflow_def = create_gnd_nextflow_params.NXF_SCRIPT
```
rbdavid: Instead of this and lines 7-11. Couldn't you do: 

```python3
if pipeline == "colorssn":
    import create_colorssn_nextflow_params as params_generator
elif pipeline == "est":
    import create_est_nextflow_params as params_generator
# and so on

# then you can just call params_generator instead of each module explicitly. 
params_file = params_generator.render_params(**vars(args)) 
workflow_def = params_generator.NXF_SCRIPT
```

You will need to rethink the `create_parser` function a bit, where the `subparsers` object might not be needed. Instead, you would just be able to do: 
```python3 
parser = argparse.ArgumentParser(description="Create a job script from a template that runs Nextflow")
params_generator.add_args(parser)
```

Since the import line happens in `__main__` before the `create_parser` function is called, the `params_generator` module name will be assigned correctly for the pipeline and everything should work.

rbdavid (approved):
Just suggested some pythonic improvements but these aren't high importance.

bin/create_nextflow_job.py
rbdavid: Ah. I'm just re-reading my suggestion. I got turned around a bit and didn't recognize that `check_args` and `create_parser` functions get called before any of the other lines in `__main__`. So my suggestion is all backwards. You'd want to implement the detection of which pipeline is having a batch script created before or within `create_parser`. Which may not be as clean as I had originally intended.

nilsoberg: I agree that this isn't clean.  There is some code reproduction which is not optimal.  This will get refactored as part of a bigger update in the future, but we don't have time to refactor this now due to the accelerated development timeline.

rbdavid (approved):
LGTM

PR #174, opened 2025-04-24
153 restrict sequences

### LINKED ISSUE #153: Restrict sequences to specified family(s) (option C and D only)

nilsoberg:
Add filter to restrict sequences in FASTA or Accession input files to a specific family(s).

### DIFF
--- a/lib/EFI/Import/Filter/Family.pm
+++ b/lib/EFI/Import/Filter/Family.pm
@@ -31,6 +31,7 @@ sub parseFamilies {
 
     my @fams = split(m/,/, uc $families);
     my $fams = {};
+    my @clans;
 
     foreach my $fam (@fams) {
         my $key;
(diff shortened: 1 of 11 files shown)


lib/EFI/Import/Filter/Family.pm
```diff
@@ -0,0 +1,94 @@
...
+    my $fams = {};
+
+    foreach my $fam (@fams) {
+        my $key;
+        if ($fam =~ m/^PF/) {
+            $key = "PFAM";
+        } elsif ($fam =~ m/^IPR/) {
+            $key = "INTERPRO";
+        } else {
+            next; # Only support Pfam and InterPro families
```
rbdavid: The web interface includes filtering by Pfam clans format (CLxxxx). Should that pattern be parsed correctly? Or is that an outdated format?

tests/modules/04l_est_accession_filter_family_fail_empty.sh
rbdavid, line 1: Just FYI, this test fails at the `cat_fasta_files` process of the est.nf workflow. Is that the correct step? 

The error that I got is: 
```
ERROR ~ Error executing process > 'cat_fasta_files'

Caused by:
  Process `cat_fasta_files` terminated with an error exit status (1)


Command executed:

  cat *.fasta > all_sequences.fasta

Command exit status:
  1

Command output:
  (empty)

Command error:
  cat: '*.fasta': No such file or directory

Work dir:
  /tmp/efi_tests/result2/04l_est_accession_filter_family_fail_empty/work/83/abef261a297c08a3689b49083f359e

Tip: you can try to figure out what's wrong by changing to the process work dir and showing the script file named `.command.sh`

 -- Check '/tmp/efi_tests/result2/04l_est_accession_filter_family_fail_empty/nextflow.log' file for details
```

rbdavid (approved):
LGTM

tests/modules/04l_est_accession_filter_family_fail_empty.sh
nilsoberg: Yes, this should fail for now

lib/EFI/Import/Filter/Family.pm
nilsoberg: Good catch.  The import supports Pfam clans (CL####), but this wasn't supported in the filtering.  This will be added.

PR #176, opened 2025-04-28
Fix blast_num_matches regression

### LINKED ISSUE #175: Fix number of blast matches null value
An issue arose during one of the last merges that caused the blast_num_matches value to be null.

nilsoberg:


rbdavid (approved):
That fixes the problem.

PR #180, opened 2025-05-07
30 implement support for domains in import

### LINKED ISSUE #30: Implement support for domains in import process
Import processes should implement functionality that can retrieve portions of sequences trimmed to domain boundaries (does not apply to Option A):

- [x] Option B: sections of the sequences that conform to the input families are retrieved
- [x] Option C: apply domain filtering to one and only one input family specified through user input
- [ ] Option D: apply domain filtering to one and only one input family specified through user input
- [ ] Domain filtering can be applied to n- or c-terminal regions

nilsoberg:


lib/EFI/Import/Config/Filter.pm
```diff
@@ -40,9 +40,9 @@ sub addImportOptions {
     $self->addOption("sequence-version=s", 0, "source sequence type (one of uniprot, uniref90, uniref50), defaults to uniprot", OPT_VALUE, "uniprot");
     $self->addOption("sequence-meta-file=s", 0, "path to the output file to save filtered sequences to", OPT_FILE);
     $self->addOption("accession-table-file=s", 0, "path to the output file to save the filtered UniRef and UniProt accession ID table to (for sunburst)", OPT_FILE);
-    $self->addOption("sequence-ids-file=s", 0, "path to the output file to save filtered sequence IDs to (for sequence retrieval)", OPT_FILE);
     $self->addOption("source-stats-file=s", 0, "path to the file containing source import stats", OPT_FILE);
     $self->addOption("stats-file=s", 0, "path to the file to save filter statistics to (appends to source stats)", OPT_FILE);
+    $self->addOption("retrieval-ids=s", 0, "path to the file to save IDs that are for retrieving, as opposed to those sequences in a user-specified FASTA", OPT_FILE);
```
rbdavid: This parameter should have `-file` suffix to match the other parameters/options.

bin/create_est_nextflow_params.py
```diff
@@ -26,6 +26,7 @@ def add_args(parser: argparse.ArgumentParser):
     common_parser.add_argument("--sequence-version", type=str, default="uniprot", choices=["uniprot", "uniref90", "uniref50"])
     common_parser.add_argument("--filter", action="append", type=str, help="Filter sequences, use multiple times to indicate filter types")
     common_parser.add_argument("--families", type=str, help="Comma-separated list of families to add")
+    common_parser.add_argument("--domain", choices=["central", "n-terminal", "c-terminal"], type=str, help="Trim sequences to domain boundaries")
```
rbdavid, line 29: Didn't we discuss limiting and/or removing the n-terminal and c-terminal options from the domain function? Should we focus the code to only accept "central" for the time being so that further development of the other cases can happen?

bin/create_est_nextflow_params.py
```diff
@@ -163,6 +165,15 @@ def render_params(output_dir, duckdb_memory_limit, duckdb_threads, fasta_shards,
...
         }
 
+    if domain is not None:
+        params |= {
+            "domain": domain
+        }
+        if import_mode == "accessions" and domain_family is not None:
+            params |= {
+                "domain_family": domain_family
+            }
```
rbdavid, line 175: Why is this check only happening for the "accessions" import mode? The FASTA import method on the website also has the text box where "domain_family" can be defined.

rbdavid: Maybe I'm just not remembering all of the decisions we made. Were we planning to remove the domain filter when the user gives sequences via the fasta file, since its assumed that the sequences are intended to be analyzed as they were given. This leaves the domain filter to only be applied on any sequences gathered from a "Protein Family Addition", correct?

lib/EFI/Import/Config/Filter.pm
```diff
@@ -78,9 +78,9 @@ sub validateOptions {
     # Output
     $opts->{sequence_meta_file} = get_default_path("sequence_meta", $outputDir) if not $opts->{sequence_meta_file};
     $opts->{accession_table_file} = get_default_path("accession_table", $outputDir) if not $opts->{accession_table_file};
-    $opts->{sequence_ids_file} = get_default_path("sequence_ids", $outputDir) if not $opts->{sequence_ids_file};
     $opts->{source_stats_file} = get_default_path("source_stats", $outputDir) if not $opts->{source_stats_file};
     $opts->{stats_file} = get_default_path("import_stats", $outputDir) if not $opts->{stats_file};
+    $opts->{retrieval_ids_file} = get_default_path("retrieval_ids", $outputDir) if not $opts->{retrieval_ids_file};
```
rbdavid, line 83: This boolean check will always fail since the option name is currently "retrieval_ids". Associated with the suggested change above.

lib/EFI/Import/SequenceDB.pm
```diff
@@ -48,34 +48,90 @@ sub getSequences {
...
+
+    open my $in, "<", $idFile or die "Unable to read input ID file '$idFile': $!";
+    open my $out, ">", $tempIdFile or die "Unable to write to temp ID file '$tempIdFile': $!";
+
+    my $domains = {};
+    my %ids;
+
+    while (my $line = <$in>) {
+        chomp($line);
+        if ($line =~ m/^([^:]+):(\d+):(\d+)$/) {
```
rbdavid, line 74: Is the `$in` file user-provided or something that we write?

rbdavid: As I understand it, the expected format is "{ID}:{start_index}:{stop_index}\n" per line. What happens if a protein is multi-domained?

lib/EFI/Import/Util.pm
```diff
@@ -97,12 +102,7 @@ B<EFI::Import::Util> is a utility module containing helpers for the various B<EF
 
 =head2 METHODS
 
-=head3 C<new()>
-
-Create a new instance of this module.
-
-
-=head3 C<batchRetrieveIds($ids, $sqlPattern, $idCol>
+=head3 C<batchRetrieveIds($ids, $sqlPattern, $idCol)>
```
rbdavid: Missing input argument.

rbdavid (commented):
There were some changes implemented in the nextflow and functional perl codes associated with fasta files. As far as I can tell, they are not associated with domains import. Could you briefly describe what those changes are for? Is it just shifting of tasks from the `import_fasta.pl` script to the `filter_ids.pl` script? 

In regards to Domains, 
- I don't see any applied limit to the number of Pfam Ids that can be given for the "domain_family" parameter (whether in python or in the perl codes). Should we implement that validation explicitly? We had discussed that that input parameter should only accept one single Pfam ID; if more or a clan or Interpro ID was given, the job should fail due to ambiguous definitions/sources of domain information. 
- Also, hadn't we decided to limit the use of n-terminus and c-terminus domain selections to protein sequences with only one domain, as a means to avoid the complex behavior of n/c terminal regions that are actually just inter-domain regions? I'm fine with leaving the code as is, just implementing some input validation or checks to prevent the complexities of multi-domained proteins and our domain filtering.

lib/EFI/Import/SequenceDB.pm
nilsoberg: `$in` is the file handle.  Multi domain proteins have different indexes.

bin/create_est_nextflow_params.py
nilsoberg: I'm leaving the code to allow all possible options.  Front end will determine functionality.

bin/create_est_nextflow_params.py
nilsoberg: If the input is a list of accessions, the only way to use a domain is to specify a family.  This only applies to accession.

nilsoberg:
I decided to keep the code as flexible as the previous codebase was.  Functionality will be determined by the user interface.

PR #181, opened 2025-05-21
74 implement support for domains in ssn creation

### LINKED ISSUE #74: Implement support for domains in SSN creation
Support creation of SSNs in `generatessn` pipeline from domain sequences.

### LINKED ISSUE #142: Support domain sequences in Color SSN and GNT pipelines
Domain sequences that are included in the EST pipeline should be supported in the Color SSN and GNT pipelines.  Currently the full sequence is output in the FASTA files generated by the pipelines, and if a sequence is a domain sequence only the domain should be included.  See issues #74 and #30.

nilsoberg:
closes #74 closes #142


lib/EFI/Import/SequenceDB.pm
```diff
@@ -71,12 +71,13 @@ sub parseIdFile {
...
-        if ($line =~ m/^([^:]+):(\d+):(\d+)$/) {
-            push @{ $domains->{$1} }, [$2, $3];
-            $out->print("$1\n") if not $ids{$1};
-            $ids{$1} = 1;
+        my ($id, @p) = split(m/:/, $line);
+        if (@p == 2) {
+            push @{ $domains->{$id} }, [@p];
+            $out->print("$id\n") if not $ids{$id};
+            $ids{$id} = 1;
         } else {
```
rbdavid, line 79: What's changed on IdFile side of things to require this change? Have I missed something key about this PR or a past PR? Or has no change in file format been implemented and this code is avoiding some other pitfall? 

Is it possible that `@p` is greater than 2, where values beyond the 1st and 2nd indices are uninteresting? Why not use `if (@p >= 2)`? Just curious.

rbdavid: The reason I bring this up: anytime I see a change in regex patterns/usage, I start to wonder how the changes might cause corner cases to arise. I don't immediately see how the previous and current methods of parsing the line differ but I also don't have a great handle on the diversity of line formats in the IdFile that is being parsed by this function.

lib/EFI/Import/SequenceDB.pm
```diff
@@ -101,7 +102,7 @@ sub convertSequences {
 
     while (my $line = <$in>) {
         chomp($line);
-        if ($line =~ m/^>(\w\w\|)?([A-Za-z0-9_\.]+)(\|.*)?$/) {
+        if ($line =~ m/^>(\w\w\|)?([A-Za-z0-9_\.]+).*?$/) {
             $curId = $2;
```
rbdavid, line 106: This is assuming a specific sequence header-line format for FASTA files but how has that format changed to require this regex change? Is it just ignoring anything beyond the "ID" portion (group 2 in the regex), just in a cleaner fashion? 

Do we have tests or validations to check for correct formatting of the input fasta file?

rbdavid: Won't this regex pattern have bugs if the line starts with only one word (\w)?

lib/EFI/Util/FASTA.pm
```diff
@@ -78,6 +51,29 @@ sub sanitize_sequence {
...
+    my $filePath = shift;
+
+    my $seqId = "";
+    my $sequences = {};
+
+    open my $fh, "<", $filePath or die "Unable to read FASTA file '$filePath': $!";
+
+    while (my $line = <$fh>) {
+        chomp $line;
+        if ($line =~ m/^>(.+)/) {
```
rbdavid, line 64: In this function, the sequence ID is the first word in the sequence header line. In `lib/EFI/Import/SequenceDB.pm`, the `convertSequences` subroutine uses a `m/^>(\w\w\|)?([A-Za-z0-9_\.]+).*?$/` regex pattern, which can match this subroutine's expected format but also a different format where the sequence ID is the third word on the line.  

What's the context/usage of the two functions?

pipelines/generatessn/create/create_full_ssn.pl
```diff
@@ -1,408 +1,255 @@
...
+# Parameters:
+#    $inputBlast - path to results from all-by-all BLAST
+#    $maxEdges - maximum number of edges to use
+#
+# Returns:
+#    1 if number of edges is valid, 0 otherwise
+#    sequence type (e.g. family domain or full)
+#    number of edges in the BLAST file (computed using the Linux 'wc' command)
+#
+sub validateInputBlast {
```
rbdavid, line 200: I guess I'm confused why the maxEdges cutoff is being applied on the raw all-to-all blast alignment file. That file has likely already had an e-value threshold cutoff applied to the results that happens during the all-to-all blast step. But, those alignment results are going to have another filter applied (alignment score, PID, etc) to remove uninteresting alignment results. Edges that pass this second filter are what get visualized by Cytoscape and define the graph of interest. Shouldn't this maxEdges check happen after _that_ filtering happens?

rbdavid: If I understand correctly, the maxEdges cutoff is purely to prevent writing SSN xgmml files that are too large to be loaded into cytoscape for a reasonable computer setup.

pipelines/generatessn/generatessn.nf
```diff
@@ -60,19 +60,11 @@ process create_full_ssn {
         path ssn_meta_file
     output:
         path "full_ssn.xgmml"
+        path "job.finish"
```
rbdavid, line 63: If the `touch job.finish` command is going to be a central mechanism to signify to a job executor that a nextflow pipeline has completed successfully, should we be using more explicit file naming? Like, `generatessn_{jobid}.finish`  instead of `job.finish`? Or is the parent directories sufficient to differentiate `job.finish` files from any of the workflows?

rbdavid (commented):
My local testing is still running. Will report back with an approve once those succeed. 

I am noticing that no new tests have been added to this PR. Are the changes made here covered by tests already implemented?

pipelines/generatessn/generatessn.nf
nilsoberg: Parent directory is sufficient

pipelines/generatessn/create/create_full_ssn.pl
nilsoberg: This is simply a sanity check (from legacy code) to prevent absurdly large SSNs from being output.

lib/EFI/Import/SequenceDB.pm
nilsoberg: Domain IDs are formatted as `UNIPROTID:#:#` and the above code wasn't parsing it correctly.  This code simply outputs the ID and saves the domain regions for use later when saving a substring of the sequence.

lib/EFI/Import/SequenceDB.pm
nilsoberg: The UniProt database IDs (from UniProt) start with tr| or sp| to indicate if they come from TrEMBL or SwissProt.  This code handles that, and corrects an issue with the end bit of the regex that arose in the custom databases we used for testing.

nilsoberg: This has been around since day one.  The code will have to support alternative IDs in the future but there's no sense solving that problem until we need to.

PR #199, opened 2025-08-19
198 refactor est pipeline script

### LINKED ISSUE #198: Refactor EST pipeline script
Clean up the EST pipeline script by refactoring the workflow into separate workflows that have logical functionality.

nilsoberg:


pipelines/est/import/import_fasta.pl
```diff
@@ -129,8 +129,8 @@ =head2 NAME
 
 =head2 SYNOPSIS
 
-    import_fasta.pl --uploaded-fasta <FILE> [--seq-mapping-file <FILE> --sequence-meta-file <FILE>
-        --output-sequence-file <FILE> --sequence-ids-file <FILE>]
```
rbdavid, line 133: `--sequence-ids-file` was removed from the accepted arguments for this script, but no changes to the source code were implemented. Was this argument depreciated already and this doc section forgotten when that change happened?

rbdavid: Or is the value for the `opts.sequence_ids_file` now just hardcoded to some value rather than possibly being set to something unsystematic?

pipelines/est/import/import_fasta.pl
```diff
@@ -176,37 +176,30 @@ =head3 Arguments
...
-
-=item C<--sequence-ids-file> (required, default value)
-
-The path to an output file that contains any IDs that were provided to this script via the metadata
-file but were not in the original FASTA file.  This occurs when a family is added to a FASTA
-import.  This file is used later in the EST pipeline to retrieve non-FASTA sequences.
-Defaults to C<sequence_ids.tab> in the current directory.
+Defaults to C<all_sequences.tab> in the current directory if not specified.
 
 =back
```
rbdavid, line 204: Should these documentation sections also describe the outputs (files, results, logging messages, etc)?

lib/EFI/Logging.pm
```diff
@@ -1,26 +0,0 @@
-
-package EFI::Logging;
```
rbdavid, line 2: Was this package unused?

lib/EFI/Config.pm
```diff
@@ -1,167 +0,0 @@
-
-package EFI::Config;
```
rbdavid, line 2: This is a lot of code to be whole-sale removing. Was this package unused?

lib/EFI/IdMapping/Builder.pm
```diff
@@ -1,236 +0,0 @@
-
-package EFI::IdMapping::Builder;
```
rbdavid, line 2: This is a lot of code being removed. Was this package not used?

lib/pyEFI/pyEFI/processing.py
```diff
@@ -0,0 +1,104 @@
+
+import pandas as pd
+
+
+def count_lengths(count_file: str, frac: float) -> pd.DataFrame:
+    """
+    Load and trim length histogram data
+
+    This function can also trim ends of the data. The method to do this
+    is borrowed from the original Perl code and it seems to include a
```
rbdavid, line 10: What does this doc string mean? Could the code's intended purpose be clarified?

lib/pyEFI/pyEFI/processing.py
```diff
@@ -0,0 +1,104 @@
...
+    Parameters
+    ----------
+        count_file
+            path to a 2 column tsv (length and count)
+        frac
+            percentage of counts to include
+
+    Returns
+    -------
+        A pandas DataFrame object with "count" and "length" columns
```
rbdavid, line 22: I think this could be clarified more. Something like: 
```
... "count" containing histogram counts (y-axis) versus the sequence "length" (x-axis). Values have been trimmed on both sides (short and long sequence lengths) based on the given `frac` value.
```

lib/pyEFI/pyEFI/processing.py
```diff
@@ -0,0 +1,104 @@
+
+import pandas as pd
+
+
+def count_lengths(count_file: str, frac: float) -> pd.DataFrame:
```
rbdavid, line 5: Should a default value be given for `frac`?

lib/pyEFI/pyEFI/processing.py
```diff
@@ -0,0 +1,104 @@
...
+    This function can also trim ends of the data. The method to do this
+    is borrowed from the original Perl code and it seems to include a
+    certain percentage of the total count.
+
+    Parameters
+    ----------
+        count_file
+            path to a 2 column tsv (length and count)
+        frac
+            percentage of counts to include
```
rbdavid, line 18: "percentage of counts to include after trimming." This would clarify how the argument value affects the final result. Also, might be worth noting the accepted value range is from 0 to 1.

lib/pyEFI/pyEFI/processing.py
```diff
@@ -0,0 +1,104 @@
...
+            minimum number of edges needed to retain a group
+
+        min_num_groups
+            keep at least this many groups (may override min_num_edges)
+
+    Returns
+    -------
+        A set of group numbers to exclude
+    """
+    sizes = sorted(group_edge_counts.itertuples(index=False))
```
rbdavid, line 60: This is an unexpected way of sorting the data array. It is highly dependent on the organization of the DataFrame, while being completely ambiguous about this fact.

Instead, you can use `group_edge_counts.sort_values(by = group_edge_counts.columns[0]).itertuples(index=False)` where the `.sort_values()` call allows you to explicitly control which column is sorted upon. In this example, I'm sorting by the zeroth column, which maps to a name stored in `group_edge_counts.columns[0]`; but we could easily sort the data frame by the "edge_count" column (I'm naive of what index that column is in the data frame).

rbdavid: Also, if you use `.itertuples(index=True)` alongside the `sort_values()` method, you keep the original indexing of rows, allowing for much cleaner recovery of row indices to keep or to cut.

rbdavid: I'm going to stop critically thinking about the python codes since that's not the intended purpose of this PR. When there's a free moment, I might come back to these and review them more thoroughly.

pipelines/shared/python/compute_length_histogram.py
```diff
@@ -0,0 +1,124 @@
...
+                # Include if the UniProt ID is in our set of valid IDs.
+                elif header in valid_ids:
+                    include_sequence = True
+
+                if include_sequence:
+                    sequence_lengths.append(len(sequence))
+
+    except FileNotFoundError:
+        print(f"Error: FASTA file not found at '{fasta_file}'", file=sys.stderr)
+        sys.exit(1)
```
rbdavid, line 84: The try-except block is re-used quite often. This could be concentrated to a `os.path.isfile()` call that raises the `FileNotFoundError` exception within a function that accepts a message string as an argument. Something like:

```
def check_file_existence(file_path: str, excpt_msg: str):
   """
   file_path needs to be the global path string 
   """
   if not os.path.isfile(file_path):
       raise FileNotFoundError(1, excpt_msg)
```
where `1` is the return code (`errno` argument for `FileNotFoundError`) and `excpt_msg` is the call specific string to print alongside the traceback text (mapping to the `strerror` argument for `FileNotFoundError`). See docs on the `OSError` parent class for the `FileNotFoundError` (exception https://docs.python.org/3/library/exceptions.html#OSError)

rbdavid: `raise FileNotFoundError` provides the traceback messaging that is silenced by the current code's `print(excpt_msg); sys.exit(1)` lines.

pipelines/shared/python/compute_length_histogram.py
```diff
@@ -0,0 +1,124 @@
...
+
+    valid_ids = parse_accession_table(accession_table, seq_type)
+
+    sequence_lengths = compute_sequence_lengths(fasta_file, valid_ids)
+
+    if not sequence_lengths:
+        print(f"Warning: No sequences were selected for histogram generation for type '{seq_type}'. Output will be empty.", file=sys.stderr)
+
+    length_counts = Counter(sequence_lengths)
```
rbdavid, line 103: Using `collections.Counter()` is an unusual way to histogram data. It works but I think the combination of all these steps could be simplified greatly by using a data analysis package such as pandas or numpy instead of the pure-python implementations in `compute_sequence_lengths()` and `main()`. But, this suggestion isn't important for this PR.

pipelines/est/subworkflows/import.nf
```diff
@@ -0,0 +1,156 @@
...
+        path 'blast_hits.tab', optional: true
+        path 'seq_mapping.tab', emit: 'seq_mapping', optional: true
+        path 'unmatched_id.tab', optional: true
+    script:
+
+    common_args = "--efi-config ${params.efi_config} --efi-db ${params.efi_db} --mode ${params.import_mode} --sequence-version ${params.sequence_version}"
+
+    family_args = ""
+    if (params.families) {
+        family_args = "--family " + params.families
```
rbdavid, line 19: Should `params.family` be written to a file instead of being an open string entry that gets called within nextflow as part of a shell script?

pipelines/est/subworkflows/import.nf
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    if (params.domain) {
+        family_args = family_args + " --domain " + params.domain
+        if (params.domain_family) {
+            family_args = family_args + " --domain-family " + params.domain_family
```
rbdavid, line 25: Similar to above. `params.domain_family` is a text box in the efi-web forms that gets pasted as a string in the `blastall` shell call that happens below. Should a file contain this parameter value?

rbdavid: Alternatively, sanitation of parameter values happens above the backend level and we can leave these parameters alone.

pipelines/est/subworkflows/import.nf
```diff
@@ -0,0 +1,156 @@
...
+    publishDir params.final_output_dir, mode: 'copy'
+    input:
+        path accession_table
+        path sequence_metadata
+    output:
+        path 'sunburst_tax.json'
+    script:
+    """
+    perl $projectDir/import/get_sunburst_data.pl --efi-config ${params.efi_config} --efi-db ${params.efi_db}
+    """
```
rbdavid, line 90: This block takes in as input `accession_table` and `sequence_metadata` but doesn't use them in the `get_sunburst_data.pl` call. Is that intended? 

The perl script has documentation where it accepts `--sequence-meta-file <FILE> --accession-table-file <FILE>` command line arguments.

pipelines/est/subworkflows/import.nf
```diff
@@ -0,0 +1,156 @@
...
+
+process import_fasta {
+    publishDir params.final_output_dir, mode: 'copy'
+    input:
+        path sequence_metadata
+        path seq_mapping
+    output:
+        path "imported_sequences.fasta", emit: "fasta_file"
+    """
+    perl $projectDir/import/import_fasta.pl --uploaded-fasta ${params.uploaded_fasta_file} --sequence-mapping-file ${seq_mapping} --output-sequence-file imported_sequences.fasta
```
rbdavid, line 119: Similar to above, `sequence_metadata` is an input to this process block but goes unused. Is that intended? There's another instance where `--sequence-meta-file` is an optional command line argument for the `import_fasta.pl` script.

pipelines/est/subworkflows/reporting.nf
```diff
@@ -0,0 +1,84 @@
...
+    input:
+        path length_histogram_file
+    output:
+        path '*.json', emit: json
+        path '*.png', emit: plots
+    """
+    base_name=\$(basename "${length_histogram_file}" .histogram.txt)
+    python $projectDir/visualization/plot_length_data.py --lengths $length_histogram_file --job-id ${params.job_id} --frac 1 --plot-filename length_histogram_\${base_name} --output-type png --proxies sm:48
+    python $projectDir/visualization/export_length_histogram_json.py --lengths $length_histogram_file --frac 1 --output-json-filename length_histogram_\${base_name}.json 
+    """
```
rbdavid, line 50: Does there need to be a check for whether the output files `*.json` and `*.png` exist before ending the process block? If the files don't exist (for whatever reason, such as no data is histogrammed), does nextflow error out because they don't exist?

rbdavid (commented):
I'll come back and review the python codes when time allows. I have some general questions about the clean up of certain files. I have one major question about the nextflow pipelines, regarding checks for existence of files before trying to output them.

lib/EFI/IdMapping/Builder.pm
nilsoberg: No longer needed

lib/EFI/Config.pm
nilsoberg: No longer needed

lib/EFI/Logging.pm
nilsoberg: No longer needed

pipelines/est/import/import_fasta.pl
nilsoberg: `--sequence-ids-file` was removed from the POD because the IDs are now loaded from the `--sequence-meta-file`.  That change was made in previous updates and not updated in the POD.

pipelines/est/import/import_fasta.pl
nilsoberg: At some point it might be good to document the file format.  Given that they are mostly files that are passed between processes in the pipeline, it isn't strictly necessary, since they aren't user-facing.

pipelines/est/subworkflows/import.nf
nilsoberg: At some point in the future we can write parameters to a config file, similar to what is done in the job manager and passed to Nextflow.  That would be outside the scope of this issue.  See issue #202

pipelines/est/subworkflows/import.nf
nilsoberg: See issue #202

pipelines/est/subworkflows/import.nf
nilsoberg: Some scripts have default inputs and outputs.  This is not consistently applied, and one or the other should be chosen.  In this case `accession_table` has a specific file name (defined in `EFI::Import::Config::Defaults`).  This should be fixed, but not part of the current issue.  See #202 and #203

pipelines/est/subworkflows/import.nf
nilsoberg: See above

pipelines/est/subworkflows/reporting.nf
nilsoberg: Nextflow will terminate with an error if the files do not exist.  I think this is the desired behavior because if they don't exist that means there was an issue earlier in the pipeline (i.e. no sequences were found).

rbdavid:
All makes sense now. LGTM

PR #207, opened 2025-08-25
192 concentrate config files into one

### LINKED ISSUE #192: Concentrate config files into one
Currently, there are nextflow config files for each EFI pipeline code and for each job executor. Instead, we could overload a single job executor-specific config file to contain information for all EFI pipelines. This would decrease automation complexity (point to a single file for all pipelines rather than dynamically choosing the config file for the pipeline) and decrease the configuration surface area that we'd need to manage.

nilsoberg:
Centralized the config files into one per platform.  This also entailed updating all of the test modules.  Additionally, assets were moved into a subdirectory of `pipelines/shared` in order to avoid mount point issues with Docker.


conf/pbspro.config
```diff
@@ -15,22 +16,24 @@ process {
...
-        memory = 32.GB
-        time = 24.h
+    withName: 'color_ssn|create_gnd|create_gnns|color_gnt_ssn|compute_clusters' {
+        memory = 10.GB
+        time = 6.h
+    }
+
+    withName: 'filter_blast|create_full_ssn' {
+        memory = 16.GB
+        time = 6.h
```
rbdavid, line 37: Since you assign a `time` value to these, are you assuming that these process blocks get submitted to the pbspro/slurm queuing system? If not, then these process blocks will be run as part of the original `nextflow run` batch script and will be subject to that batch script's allocation and wall-clock limit. If they are supposed to be submitted to the queue, you need to include the `executor = 'pbspro'` line.

tests/modules/extended/01c_est_family_filter_taxonomy_default_path.sh
```diff
@@ -3,8 +3,6 @@ set -e
 
 TEST_RESULTS_DIR=$1
 CONFIG_FILE=$2
-NXF_EST_CONFIG_FILE="conf/est/$CONFIG_FILE"
-NXF_SSN_CONFIG_FILE="conf/generatessn/$CONFIG_FILE"
```
rbdavid, line 7: I actually prefer these changes, where you remove the two lines defining variables that point to the same file, over the changes applied in the core tests. Could the core tests be made to match these tests? Or does that raise ambiguity that we want to avoid?

tests/runtests.sh
```diff
@@ -16,7 +16,7 @@ if [[ ! -e pipelines/generatessn/generatessn.nf || ! -e pipelines/est/est.nf ]];
 fi
 
 if [[ $# -ne 1 ]]; then
-    CONFIG_FILE="docker.config"
+    CONFIG_FILE="conf/docker.config"
 else
     CONFIG_FILE=$1
```
rbdavid, line 21: We'll need to update documentation to specify that the first argument needs to be the relative or global path to the `conf/docker.config` file rather than the `docker.config` that we currently say. Or just change line 21 to also include the `conf/` dir too.

rbdavid (commented):
Some minor comments about specific conf files and the runtests.sh script. Other than those, looks good to me.

conf/pbspro.config
nilsoberg: I think this file should be considered an example.  The exact time requirements will depend on the cluster that this runs on.  Perhaps this file should be renamed to `pbspro.config.example` and the documentation updated in the testing guide indicating that the user must copy the file to `pbspro.config` and update as necessary? @rbdavid

tests/modules/extended/01c_est_family_filter_taxonomy_default_path.sh
nilsoberg: Yes, I agree.

rbdavid (approved):

conf/pbspro.config
rbdavid: Just to clarify, I don't have a problem with the `time` keyword being there. But for this to be a complete example of a `withName: process_block` section that gets submitted as a separate batch script, we should explicitly include the `executor = 'pbspro'` lines. Otherwise, nextflow assumes the "local" or default executor is used.

Not an important change in the grand scheme of things, since these conf files are examples that need to be adapted to the compute resource.

Issue #208, opened 2025-08-26
Properly apply fraction filter


nilsoberg:
Currently the fraction filter is applied to all input sequences after they are collected.  Previous functionality was to apply the fraction only to the family input.  Update the code so that the fraction is applied only to those sequences that originate from a family only.  The metadata table has an entry that indicates where the sequence came from.

nilsoberg:
close as duplicate of #166

PR #209, opened 2025-09-03
111 support uniref sequences throughout pipelines

### LINKED ISSUE #111: Support UniRef sequences throughout pipelines

nilsoberg:
* Fix missing Pfam no_neighbors output IDs
* Skip Pfam table output for clusters that are singletons
* Properly expand metanodes/repnodes to support UniRef in GNT - gnt pipeline updated to use metanode mapping that is output from shared colorssn pipeline
* Fix `EFI::GNT::Neighborhood` query code that returned excessive number of Pfam families, causing column truncation and missing data
* Fix Color SSN output caused by missing parameters to a function in `EFI::SSN::XgmmlReader`
* Remove usage of some strings by using shared constants (e.g. `SEQ_*` from `EFI::Sequence::Type`)
* Refactor Xgmml writing code to remove code that was duplicated; code and POD from `EFI::GNT::GNN::XgmmlWriter` was moved to `EFI::Xgmml::Writer`
* Update rST due to POD updates (include from unrelated code -- byproduct of running build process)
* Update POD
* Update tests
* Did not complete support for UniRef in GND pipeline -- more complicated than the scope of this issue


docs/source/pipelines/est/import/filter_ids.rst
```diff
@@ -20,13 +20,160 @@ Usage
...
+
+
+Example Usage
+^^^^^^^^^^^^^
+
+::
+
+   filter_ids.pl --filter fraction=10 ...
+
+   # If 200 sequences are provided, then 20 will be output
```
rbdavid, line 107: This will need to be updated when issue #208 is added.

docs/source/pipelines/est/import/import_fasta.rst
rbdavid, line 1: Should `--sequence-ids-file` be in this documentation? Its listed on lines 10 and 21 but removed from line 42 and its description is removed as well.

lib/EFI/GNT/GNN/Hubs.pm
```diff
@@ -227,13 +227,19 @@ sub computeClusterHubs {
                 foreach my $queryId (keys %$pfamHub) {
                     map { $unclassified{$_->{id}} = 1 } @{ $pfamHub->{$queryId} };
                 }
-                next;
```
rbdavid, line 230: Removal of this line will result in "NONE_PFAM" labeled neighbors being counted in the GNN as a spoke node, correct? Is that intended?

rbdavid: Will this just be a singular hub node labeled "NONE_PFAM"?

lib/EFI/GNT/GNN/XgmmlWriter.pm
```diff
@@ -15,37 +15,46 @@ use EFI::Annotations;
...
+# writeStarting - protected
+#
+# Write the starting tags, e.g. graph.
+#
+sub writeStarting {
     my $self = shift;
+    my %attr = @_;
 
-    $self->{output} = IO::File->new(">" . $self->{output_file});
+    $attr{label} = "GNN" if not $attr{label};
```
rbdavid, line 43: Why not use `//` defined-or operator as done in line 28?

lib/EFI/GNT/GNN/XgmmlWriter/PfamHub.pm
```diff
@@ -53,16 +60,21 @@ sub write {
...
+
+            $hasSpoke = 1;
         }
 
         # Only write a hub node if there were spoke nodes
-        if (@clusterNums) {
+        if ($hasSpoke) {
             my $nodeAttr = $self->getHubData($pfamHubName, $hub);
             my ($familyNames, $pfamShortName, $pfamLongName) = $self->{gnt_anno}->getFamilyNames($pfamHubName);
+            $pfamShortName = NONE_PFAM if not $pfamShortName;
```
rbdavid, line 71: The short name will just be "none"? Why not something more descriptive, like "Unlabeled"?

lib/EFI/SSN/XgmmlReader/IdList.pm
```diff
@@ -25,7 +26,7 @@ sub new {
     $self->{id_list_fields} = { map { $attrDisplay->{$_} => $_ } @$attrNames };
     $self->{metadata} = {}; # any node metadata that we are to store (e.g. swissprot)
     $self->{meta_map} = undef; # Metanode -> IDs in metanode mapping
-    $self->{id_type} = "uniprot";
+    $self->{id_type} = SEQ_UNIPROT;
```
rbdavid, line 29: When this `new()` subroutine is called in ssn_to_id_list.pl, this line hardcodes the `id_type` variable to map to "uniprot". How are UniRef90/50 sequence id types enabled/usable with this code?

lib/EFI/SSN/XgmmlReader/IdList.pm
```diff
@@ -49,10 +50,10 @@ sub getMetanodeSizes {
...
     my $self = shift;
     my $idf = $self->{id_type};
-    if ($idf ne "uniprot") {
-        $idf = "repnode" if $idf eq FIELD_REPNODE_IDS;
-        $idf = "uniref90" if $idf eq FIELD_UNIREF90_IDS;
-        $idf = "uniref50" if $idf eq FIELD_UNIREF50_IDS;
+    if ($idf ne SEQ_UNIPROT) {
+        $idf = SEQ_REPNODE if $idf eq FIELD_REPNODE_IDS;
+        $idf = SEQ_UNIREF90 if $idf eq FIELD_UNIREF90_IDS;
+        $idf = SEQ_UNIREF50 if $idf eq FIELD_UNIREF50_IDS;
```
rbdavid: As above's comment, when is `$idf` ever not going to equate to `SEQ_UNIPROT`?

lib/EFI/SSN/XgmmlReader/IdList.pm
```diff
@@ -61,7 +62,7 @@ sub getMetanodeType {
 sub getMetanodes {
     my $self = shift;
     my $fullIds = {};
-    if ($self->{id_type} ne "uniprot") {
+    if ($self->{id_type} ne SEQ_UNIPROT) {
```
rbdavid, line 65: and again.

lib/EFI/Sequence/Type.pm
```diff
@@ -146,6 +148,28 @@ For UniRef50 (C<uniref50>) ID types.
...
+For UniProt (C<uniprot>) ID types.
+
+=item C<SEQ_UNIREF50>
+
+For UniRef50 (C<uniref50>) ID types.
+
+=item C<SEQ_UNIREF90>
+
+For UniRef90 (C<uniref90>) ID types.
+
```
rbdavid: redundant documentation text. Same content as lines 141 to 150.

pipelines/gnt/create_gnns.pl
```diff
@@ -143,18 +156,18 @@ =head2 NAME
...
 
 =head2 DESCRIPTION
 
 C<create_gnns.pl> reads a list of sequences and corresponding cluster numbers and
-creates XGMML files for a cluster GNN and Pfam GNN. It optionally can create tables
-and metadata with data about the Pfams of neighbors in the input IDs and a genome
-neighborhood diagram (GND) file.
+creates XGMML files for a cluster GNN and Pfam GNN.  It optionally can create
+tables and metadata with data about the Pfams of neighbors in the input IDs and a
+genome neighborhood diagram (GND) file.  The input IDs are assumed to be UniProt.
```
rbdavid, line 170: This might explain the confusion I was having with the lib/EFI/SSN/XgmmlReader/IdList.pm file and others.

tests/modules/07_colorssn_uniref50.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -e
+
+if [[ ! -e "$EFI_TEST_SSN_UNIREF50" ]]; then
+    echo "Test skipped; missing $EFI_TEST_SSN_UNIREF90"
```
rbdavid: Should be `$EFI_TEST_SSN_UNIREF50`

rbdavid (approved):
Generally, looks good to me. I just had a few clarifications/questions.

docs/source/pipelines/est/import/import_fasta.rst
nilsoberg: rST is out of date, fixed now

lib/EFI/GNT/GNN/Hubs.pm
nilsoberg: The code was "wrong" in that it resulted in neighbor sequence IDs not being included in the output tables when an ID in the cluster was "valid".  "Wrong" is subjective; this change mimics the behavior of the official code.  The result of this change will be additional neighbors in the output tables.

lib/EFI/GNT/GNN/XgmmlWriter/PfamHub.pm
nilsoberg: Good point, it's just "none" to make it mirror what the current code does.  Would prefer to leave it matching the legacy code even if a new name would be better.

lib/EFI/GNT/GNN/XgmmlWriter.pm
nilsoberg: I generally use `//` when assigning a different variable, but it could be done either way.  I think using the `if` syntax is a bit more idiomatic in the sense that we don't do a reassignment of the same value in the case that `attr` is defined.  `title` on line 28 always has to be assigned, but `label` only needs to be assigned if it doesn't already exist.  I agree that `//` is shorter but my practice generally follows what I wrote.

lib/EFI/Sequence/Type.pm
nilsoberg: Good catch

lib/EFI/SSN/XgmmlReader/IdList.pm
nilsoberg: The code assumes that a SSN is UniProt by default since the definitive test is whether the `EFI::Annotations::Fields::FIELD_UNIREFXX_IDS` node attribute is present.  The `id_type` is set the first time a UniRef node attribute is encountered in `processNodeAttribute`.

lib/EFI/SSN/XgmmlReader/IdList.pm
nilsoberg: As mentioned above, this is set in `processNodeAttribute` since that's the only definitive test for a UniRef network.

rbdavid (approved):

PR #224, opened 2025-11-21
223 Add domain_region parameter to EST nextflow pipeline

### LINKED ISSUE #223: Add domain_region to est.nf params
The pipelines/est/subworkflows/import.nf code uses params.domain in two places, once as a boolean and the other as an input argument to `pipelines/est/import/get_sequence_ids.pl` script. This in disagreement with the symfony frontend where the Job table has `domain` and `domainRegion` columns, the former mapping to a boolean while the latter is the string. So, need to add the `params.domain_region` parameter to `pipelines/est/subworkflows/import.nf` and update the python code to generate the params.json and testing suite.

rbdavid:
Add the domain_region nextflow parameter to limit the influence of the params.domain parameter to just be a boolean. This mirrors how the efi-web domain-related columns in the Job table is handled. 

closes #223


nilsoberg (approved):

rbdavid:
#231 is now created to improve the domain filtering capability. Otherwise, this PR is running domain filtering jobs as intended.

PR #226, opened 2025-12-02
Remove uniref_version dependency from generatessn pipeline

### LINKED ISSUE #218: UniRef version input argument issue
https://github.com/EnzymeFunctionInitiative/EST/blob/a805a68b97fa3cc2491b2942b55273fbeb73f9c7/pipelines/generatessn/generatessn.nf#L51

Check that the perl code is correctly implementing logic for handling uniprot, uniref90, and uniref50 jobs.

Ideally the `get_annotations.pl` script would accept the "uniprot", "uniref90", or "uniref50" strings instead of the "50" or "90" as it currently does. But, maybe this input argument isn't needed at all with the new method of handling metadata files?

nilsoberg:


rbdavid (commented):
Looks good to me. I'll run the relevant testing modules in the morning. Also, the `bin/create_generatessn_nextflow_params.py` should be updated to remove that argument. I can get to that in the morning as well.

rbdavid (approved):
Tests are passing

PR #228, opened 2025-12-04
227 update gnd color assignment

### LINKED ISSUE #227: Update GND color assignment
Update GND color assignment

- [x] Merge similar colors
- [x] Remove colors that resemble red
- [x] Families that are shared with query gene are assigned same color as query gene
- [x] Query gene is uniquely and consistently colored

nilsoberg:
Updated the GND color assignments to ensure that families that share query gene families are colored the same as query gene.  Also fixed two cluster numbering issues: first cluster was numbered `0` not `1`, and `cluster_num` was not being inserted into GND database `attributes` table.


rbdavid (approved):
I haven't run any tests on this code. But the updates make sense.

PR #239, opened 2025-12-16
237 Add implementation inspecific config file

### LINKED ISSUE #237: Create an non-implementation specific config file
Nextflow enables one config file to include settings from another config file. I'm suggesting we have a config file that contains implementation-independent information, e.g. default values for certain parameters within workflows.

The documentation for 24.04.XX can be found here: https://github.com/nextflow-io/nextflow/blob/v24.10.9/docs/config.md#includes

This implementation would simplify the interface between the backend code and frontend web forms; certain nextflow parameters are undeclared by the web forms, and so, require hard-coding their values at that interface so that the backend code can run successfully.

rbdavid:
`default_params.config` will contain a `params{}` block that lists required parameters for all EFI tools and their default values. Since its a config file read in via the `-C` cli argument to `nextflow run`, the `params.json` file takes precedence.


rbdavid:
Tests are running on my local desktop to verify that no parameters were forgotten. Otherwise, this PR is ready for review.

nilsoberg (changes requested):
This change is an improvement and I like it.  Is includeConfig a Nextflow idiom?  Also, update the additional comments to start with an uppercase letter (the EST repo attempts to keep with this standard).

rbdavid:
https://github.com/nextflow-io/nextflow/blob/v24.10.9/docs/config.md#includes is the documentation on the nextflow config keyword. This is also used in 25.X.X version of nextflow so won't cause any conflicts if/when we update to use the latest NF version.

rbdavid:
Merging after conversation with Nils.

PR #241, opened 2025-12-16
225 make single input file param for est

### LINKED ISSUE #225: Make a single nf param for the various input files for EST import modes
Since the `params.import_mode` is already used to differentiate the various inputs used during `get_sequence_ids()` nextflow block, the input file for the BLAST, FASTA, and ACCESSION input paths should be called using a uniform nextflow parameter name instead of `params.blast_query_file`, `params.uploaded_fasta_file`, or `params.accessions_file` respectively. Differentiating these three file parameters is unnecessary because only one import mode's flowchart will be traversed by a job; there's no need for further differentiation of the paths.

This would minimize the required logic in the bin/create_params... python scripts, the job handler in the job_manager code, as well as KBase.

rbdavid:
Branches off of #239. Implements the single parameter used to cover the input files for the BLAST, ACCESSIONS, and FASTA input modes for the est.nf workflow. This cleans up the parameter namespace but will require updates on KBase and job_manager code bases.


rbdavid:
This branch got rebased because I merged #239 into nextflow-test branch, making a load of conflicts. I rebased instead of merged/cherry-picked conflicts. I screwed things up here.

PR #249, opened 2026-01-15
Name SSN file according to job name

### LINKED ISSUE #233: Name output files according to job name
When creating a SSN, name the .xgmml file based on the input job name before zipping up for transferring

nilsoberg:


pipelines/generatessn/generatessn.nf
```diff
@@ -59,10 +59,18 @@ process create_full_ssn {
...
+        .replaceAll(/[^a-zA-Z0-9_.]/, "-")
+        .replaceAll(/^-+|-+$/, "")
+        .toLowerCase()
+    def file_name = "${clean_job_name}.xgmml"
+
     """
-    perl $projectDir/create/create_full_ssn.pl --blast $filtered_blast --fasta $filtered_fasta --metadata $ssn_meta_file --output full_ssn.xgmml  --title ${params.job_name} --db-version ${params.db_version} --stats stats.json
-    zip full_ssn.xgmml.zip full_ssn.xgmml
-    rm full_ssn.xgmml
+    perl $projectDir/create/create_full_ssn.pl --blast $filtered_blast --fasta $filtered_fasta --metadata $ssn_meta_file --output ${file_name} --title "${params.job_name}" --db-version ${params.db_version} --stats stats.json
```
rbdavid: `--output "${file_name}"` would be the safe thing to do, right? As you've done with the `--title` argument.

pipelines/generatessn/generatessn.nf
```diff
@@ -59,10 +59,18 @@ process create_full_ssn {
...
+    zip full_ssn.xgmml.zip ${file_name}
+    rm ${file_name}
```
rbdavid: See above.

rbdavid (commented):
What happens when no characters remain after the `.replaceAll()` lines? The file could possibly just be `.xgmml`. 

Won't efi-web code sanitize the job_name string? The job_manager code makes no requirements on a Job's `job_name` attribute. If a local installation of EST wants to use non-ASCII or non-english characters, shouldn't they be able to? I guess I'm arguing for not implementing string sanitization in the backend code.

Also, I want to note that this code will now fail if `job_name = null`, which is different behavior than the previous code. Not an issue, just something to note.

Finally, the associated issue for this PR mentions making the network name informative when visualized in Cytoscape. Have we already fixed that part of the issue? Is that what the `--title` argument is for the `perl .../create_full_ssn.pl` script?

PR #255, opened 2026-02-02
238 fix demultiplex process

### LINKED ISSUE #238: Fix demultiplex process
```
ERROR ~ Error executing process > 'ALL_BY_ALL:demultiplex'

Caused by:
  Process `ALL_BY_ALL:demultiplex` terminated with an error exit status (1)


Command executed:

  echo "COPY (SELECT * FROM read_parquet('1.out.parquet')) TO 'mux.out' (FORMAT CSV, DELIMITER '	', HEADER false);" | duckdb
  perl /home/rbdavid/Scripts/git/EST/pipelines/est/mux/demux.pl -blastin mux.out -blastout 1.out -cluster sequences.fasta.clstr
  python /home/rbdavid/Scripts/git/EST/pipelines/est/mux/transcode_demuxed_blast.py --blast-output 1.out

Command exit status:
  1

Command output:
  Read in clusters
  Demultiplex blast

Command error:
  Read in clusters
  Demultiplex blast
  Traceback (most recent call last):
    File "/home/rbdavid/Scripts/git/EST/pipelines/est/mux/transcode_demuxed_blast.py", line 55, in <module>
      csv_to_parquet(blast_output, read_options, parse_options, convert_options)
    File "/usr/local/lib/python3.10/site-packages/pyEFI/transcode.py", line 42, in csv_to_parquet
      writer.write_batch(batch)
    File "/usr/local/lib/python3.10/site-packages/pyarrow/parquet/core.py", line 1078, in write_batch
      self.write_table(table, row_group_size)
    File "/usr/local/lib/python3.10/site-packages/pyarrow/parquet/core.py", line 1102, in write_table
      raise ValueError(msg)
  ValueError: Table schema does not match schema used to create file: 
  table:
  qseqid: string
  sseqid: string
  pident: float
  alignment_length: uint32
  bitscore: float
  query_length: uint32
  subject_length: uint32
  alignment_score: uint32 vs. 
  file:

Work dir:
  /tmp/est_testing/results2/01a_est_family/work/83/3ee01ccf18ab882db3f90b94df3fe1

Tip: when you have fixed the problem you can continue the execution adding the option `-resume` to the run command line

 -- Check '/tmp/est_testing/results2/01a_est_family/nextflow.log' file for details
```

nilsoberg:
These updates fix the original demultiplex issue in addition to the following changes:

- Fix Python scripts to add missing columns to schema when converting CSV to parquet
- Convert Perl demux code to Python
- Fix pipeline [use copy of original fasta when multiplexing so that the original wasn't modified and used in later steps (convergence ratio calculation)]
- Write unit test (new `pipelines/verify` directory and supporting bash scripts) to ensure Python code functions like Perl code
- Rename multiplex/demultiplex to collapse/expand to reflect actual computational purpose
- Update docs to reflect naming

There were two problems: The first was that an empty list of columns was being passed to a function that converted CSV to parquet.  This resulted in a parquet file that had no columns and the conversion wasn't possible because of the schema mismatch.  The second issue was that the convergence ratio was being computed wrong; one of the files passed to that program was based on the multiplexed results (when the input should have been the demultiplexed results).

NOTE: this **does not update the keys in the `params.yaml` file** that is generated from `bin/create_est_nextflow_params.py` (although the command line argument has been updated), or the `params.yaml` file that is passed from tests or job manager.  The internal key is still `multiplex`.  Changes to this flag naming need to be coordinated with future updates to the job manager.


docs/source/pipelines/est/cluster/expansion.rst
```diff
@@ -0,0 +1,15 @@
+Expanding Clustered (Redundant) Sequences)
```
rbdavid: Extra close-parentheses?

docs/source/pipelines/est/cluster/expansion.rst
```diff
@@ -0,0 +1,15 @@
...
+==========================================
+
+Clustered expansion only runs when the ``cluster`` parameter was set to true. It
```
rbdavid: Is the planned name for the new parameter to be `cluster` instead of `multiplex` or `collapse_redundancy`?

bin/create_est_nextflow_params.py
```diff
@@ -20,7 +20,7 @@ def add_args(parser: argparse.ArgumentParser):
     common_parser.add_argument("--fasta-shards", type=int, help="Number of files to split FASTA into. File is split so that BLAST can be parallelized")
     common_parser.add_argument("--accession-shards", type=int, help="Number of files to split Accessions list into. File is split so that sequence retrieval can be parallelized")
     common_parser.add_argument("--fasta-db", type=str, required=True, help="FASTA file or BLAST database to retrieve sequences from")
-    common_parser.add_argument("--multiplex", action="store_true", help="Use CD-HIT to reduce the number of sequences used in analysis")
+    common_parser.add_argument("--collapse-redundancy", action="store_true", help="Use CD-HIT to reduce the number of sequences used in analysis")
```
rbdavid: This should be `--cluster`, no? To follow the documentation being pushed in this PR.

docs/source/pipelines/est/index.rst
```diff
@@ -134,7 +134,7 @@ swapping results to disk if it is memory constrained but some operations
 require some minimum amount of memory.  If these solutions did not solve the
 problem, try using the newest version of DuckDB (which may require manually
 building the docker image) or decreasing the number of ``--blast-num-matches``
-which will reduce the total number of edges processed.  Multiplexing will also
+which will reduce the total number of edges processed.  Collapsing redundancy will also
```
rbdavid: We're jumping back and forth between "clustering" and "collapsing redundancy".

docs/source/pipelines/gnt/create_gnns.rst
rbdavid, line 1: Not sure why this is in this PR. But looks good to me.

pipelines/est/condense/restore_condensed_sequences.py
```diff
@@ -0,0 +1,105 @@
+import sys
+import argparse
+import re
+
+def parse_cdhit_clstr(clstr_path):
```
rbdavid, line 5: Add annotation details. e.g. `def parse_cdhit_clstr(clstr_path: str) -> Dict[str,List[str]]:`

You'll need to add `from typing import Dict, List` in the preamble.

pipelines/est/condense/restore_condensed_sequences.py
```diff
@@ -0,0 +1,105 @@
...
+                    if line.endswith('*'):
+                        current_rep = seq_id
+
+        # Catch the last cluster
+        if current_rep and buffer_children:
+            clusters[current_rep] = buffer_children
+            
+    return clusters
+
+def expand_edges(blastin_path, blastout_path, clusters):
```
rbdavid, line 45: Annotations would help here. I'm assuming `clusters` is the dict object output from the `parse_cdhit_clstr()` func.

pipelines/est/condense/restore_condensed_sequences.py
```diff
@@ -0,0 +1,105 @@
+import sys
+import argparse
+import re
+
+def parse_cdhit_clstr(clstr_path):
+    """
+    Parses a CD-HIT .clstr file to map representative IDs to their list of children.
+    Returns a dictionary: {representative_id: [child_id_1, child_id_2, ...]}
```
rbdavid, line 8: Also, the description should be: 
```
"""
Returns a dictionary: {representative_id: [representative_id, child_id_1, child_id_2, ...]}
"""
```

rbdavid: The doc string should probably be this to follow pyEFI doc string standards:

    """
    Parses a CD-HIT .clstr file to map representative IDs to their list of children.

    Parameters
    ----------
        clstr_path
            global or local path to the cd-hit clstr file.

    Returns
    -------
        cluster tree dictionary, with structure:
            {
                representative_id1: [
                    representative_id1, child_id_1, child_id_2, ...
                ],
                representative_id2: [
                    representative_id2, ...
                ],
                ...
            }

    """

pipelines/est/cluster/expand_clusters.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # Catch the last cluster
+        if current_rep and buffer_children:
+            clusters[current_rep] = buffer_children
+            
+    return clusters
+
+def expand_edges(blastin_path, blastout_path, clusters):
+    """
+    Reads BLAST file, looks up children for query/subject, and expands edges.
```
rbdavid: Doc string should be:


    """
    Reads BLAST file, looks up children for query/subject, and expands edges. Writes expanded BLAST results to file.

    Parameters
    ----------
        blastin_path
            global or local path to the blast output file for sequence alignments between cluster representative sequences. These results are to be expanded to represent all clusters. This input file is expected to be a tab-separated table of `blastp` results.

        blastout_path
            global or local path used to write the expanded blast output to file.

        clusters
            dictionary of cluster tree, created from `parse_cdhit_clstr()` function.
    """

pipelines/est/condense/restore_condensed_sequences.py
```diff
@@ -0,0 +1,105 @@
...
+                continue
+                
+            query_rep = parts[0]
+            subject_rep = parts[1]
+            # Keep the rest of the BLAST columns (pident, evalue, etc.)
+            rest_of_line = parts[2:] 
+
+            # Validation
+            if query_rep not in clusters:
+                print(f"SOURCE {query_rep} does not exist in the cluster file", file=sys.stderr)
```
rbdavid, line 62: I think the printed text is a bit inaccurate here. The if statement is checking whether `query_rep` is in the keys of the `clusters` dict. So, the print statement should read:

`f"SOURCE {query_rep} is not a cluster representative sequence."`

pipelines/est/condense/restore_condensed_sequences.py
```diff
@@ -0,0 +1,105 @@
...
+                continue
```
rbdavid, line 63: A big-picture note:
This boolean should never test True though since `cd-hit` outputs a fasta file that only contains rep-sequences. Why is the "validation" of the `cd-hit` results being done _after_ the all-by-all calculations have already ran?

pipelines/est/condense/restore_condensed_sequences.py
```diff
@@ -0,0 +1,105 @@
...
+                        child_s = query_children[j]
+                        # Join and write
+                        out_line = "\t".join([child_q, child_s] + rest_of_line)
+                        fout.write(out_line + "\n")
+
+            # CASE B: Cross-Hit (Cluster A hits Cluster B)
+            else:
+                if subject_rep not in clusters:
+                    print(f"TARGET {subject_rep} does not exist in the cluster file", file=sys.stderr)
+                    continue
```
rbdavid, line 83: See above big picture note.

pipelines/est/cluster/expand_clusters.py
```diff
@@ -0,0 +1,105 @@
...
+
+            # Validation
+            if query_rep not in clusters:
+                print(f"SOURCE {query_rep} does not exist in the cluster file", file=sys.stderr)
+                continue
+            
+            query_children = clusters[query_rep]
+            
+            # CASE A: Self-Hit (Cluster hits itself)
+            # This creates the internal clique (A-B, A-C, B-C) but excludes self-loops (A-A).
```
rbdavid: This comment is somewhat confusing.

I think it should read something like:
`# This creates the internal clique (A0-A1, A0-A2, A1-A2, ...) but excludes self-loops (A0-A0).`

pipelines/est/condense/restore_condensed_sequences.py
```diff
@@ -0,0 +1,105 @@
...
+            # This creates the internal clique (A-B, A-C, B-C) but excludes self-loops (A-A).
+            if query_rep == subject_rep:
+                n = len(query_children)
+                for i in range(n):
+                    for j in range(i + 1, n):
+                        child_q = query_children[i]
+                        child_s = query_children[j]
+                        # Join and write
+                        out_line = "\t".join([child_q, child_s] + rest_of_line)
+                        fout.write(out_line + "\n")
```
rbdavid, line 77: This nested for loop should be simplified to: 

```
combos = itertools.combinations(query_children, 2)
for child_q, child_s in combos:
    out_line = "\t".join([child_q, child_s] + rest_of_line)
    fout.write(out_line + "\n")
```

Will need to add itertools to preamble: `import itertools`. 
The `itertools.combinations()` function explicitly avoids self-loops (`j == i`) and can grab all pairs of members in query_children very efficiently, using optimized C code instead of two slow python for-loops. It is also memory efficient because the `combos` object is a lazy iterator (but so are your `range()` calls).

pipelines/est/condense/restore_condensed_sequences.py
```diff
@@ -0,0 +1,105 @@
...
+                if subject_rep not in clusters:
+                    print(f"TARGET {subject_rep} does not exist in the cluster file", file=sys.stderr)
+                    continue
+
+                subject_children = clusters[subject_rep]
+                
+                for child_q in query_children:
+                    for child_s in subject_children:
+                        out_line = "\t".join([child_q, child_s] + rest_of_line)
+                        fout.write(out_line + "\n")
```
rbdavid, line 90: As above, itertools will be much more efficient than this nested for-loop. 

```
combos = itertools.product(query_children, subject_children)
for child_q, child_s in combos:
    out_line = "\t".join([child_q, child_s] + rest_of_line)
    fout.write(out_line + "\n")
```

pipelines/est/subworkflows/all_by_all.nf
```diff
@@ -68,47 +68,60 @@ process blastreduce {
...
+        // All-by-all BLAST
+        fasta_shards = split_fasta(blast_input_fasta)
+        blast_fractions = all_by_all_blast(
+            blastdb.database_files,
+            blastdb.database_name,
+            fasta_shards.flatten()
+        ) | collect
+
+        // Eliminate duplicate and self-edges
         reduced_blast_parquet = blastreduce(blast_fractions, fasta_lengths_parquet)
```
rbdavid, line 116: This is a major, glaring issue. By this point in the workflow, assuming `params.multiplex = True`:
 
- We've clustered sequences down into a fasta file of only representative sequences. 
- We've then run all-by-all alignments on those representative sequences. 
- Then this `blastreduce` step removes lines in the blast result file, specifically for self-alignments (e.g. a representative sequence aligned against itself). !!! This is the big main issue. And I have a smaller issue to raise as well.

The big main issue: The original `demux.pl` and new `expand_clusters.py` codes _*expect*_ that self-matches are present so that the cluster children have alignment results to other children in the same cluster (line 67 in `expand_clusters.py`)! We cannot be removing the self-alignment results during the `blastreduce` process! Otherwise, the SSN file will have no edge drawn between sequence nodes that are 100% sequence identical and are likely the most "similar" in function! That's just wrong.

rbdavid: The smaller but, I think important, issue is: two blastp alignments between two sequences will return different results when sequence 1 is the query or when sequence 2 is the query: 

| qseqid | sseqid | pident | alignment_length | ... | evalue | bitscore |
| ------ | ------ | ------ | ---------------- | --- | ------ | -------- |
| seq1  | seq2  | 24.02 | 358 | ... | 6e-08 | 57.0 |
| seq2  | seq1  | 24.02 | 358 | ... | 4e-08 | 57.8 |

These two lines are pulled from a real blastp run. Note the two different e-value and bitscore values; these differences arise due to query sequence length being used in the numerical calculations of those metrics. The `blastreduce` README doc says that one of these lines is removed, based solely on the alphabetical ordering of `qseqid` and `sseqid` values. This seems like the wrong choice to make. I don't know what the original v1.0 perl code did to remove these "duplicates" lines but I think we should review that code before using alphabetical ordering as the method to choose winners...

pipelines/shared/nextflow/sequence.nf
```diff
@@ -29,14 +29,15 @@ process get_sequences {
...
+// Formerly known as multiplex
+process cluster {
     input:
         path fasta_file
     output:
         path "sequences.fasta", emit: "fasta_file"
         path "sequences.fasta.clstr", emit: "clusters"
     """
-    cd-hit -d 0  -c 1 -s 1 -i $fasta_file -o sequences.fasta -M 10000
+    cd-hit -d 0  -c 1 -s 1 -i ${fasta_file} -o sequences.fasta -M 10000
```
rbdavid: Should `${fasta_file}` be surrounded by quotation marks?

rbdavid (commented):
Okay, the PR looks great. I haven't tested the new verify code; do I just need to run the same testing suite that I've always done to run the verify codes/tests? 

__But__ I now see, for the first time, a major issue with the `blastreduce` process and how it interacts with this demux/cluster_expand code. Those issues should probably be their own high-priority issue and PR. Let me know what you think!

pipelines/shared/nextflow/sequence.nf
nilsoberg: Nextflow handles ensuring file names are passed properly (e.g. "staging").

pipelines/est/subworkflows/all_by_all.nf
nilsoberg: If you haven't already done the computations, could you provide example data for the `blastreduce` issue you mentioned as well as a comparison to the legacy tools?  We should manually verify that the output from the all-by-all blast has self-matches, and the `blastreduce` step does indeed (incorrectly) remove self-loops in the new process.

rbdavid: I've made an issue for this (#258). The glaring issue shouldn't hold back this PR's code, right?

rbdavid (approved):

PR #256, opened 2026-02-05
212 support blast and fasta inputs tognd pipeline

### LINKED ISSUE #212: Support BLAST and FASTA inputs to`gnd` pipeline

nilsoberg:
Refactor the GND workflow pipeline to support BLAST and FASTA input options.  This required minor updates to the EST pipeline so that a script (`get_sequence_ids.pl`) could be shared between pipelines.  New code was added to existing utility modules to assist in transforming sequence IDs from accession, FASTA, and BLAST into the proper ID format needed to support GND creation.  A minor update was made to `create_gnns.pl` in the GNT pipeline to refactor a function (into `EFI::IdMapping1) so code could be shared between GND and GNT.  About 1/2 of the updates are related to documentation.  Tests were added for testing UniRef GND functionality.


bin/create_gnd_nextflow_params.py
```diff
@@ -30,9 +36,14 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
     else:
         args = validated_args
 
-    if not os.path.exists(args.cluster_id_map):
-        print(f"SSN Input file '{args.cluster_id_map}' does not exist")
+    if not os.path.exists(args.input_file):
+        print(f"SSN Input file '{args.input_file}' does not exist")
```
rbdavid: This isn't looking for an SSN input file, so the print statement should be updated.

bin/create_gnd_nextflow_params.py
```diff
@@ -47,22 +58,29 @@ def check_args(args: argparse.Namespace) -> argparse.Namespace:
         print("Failed to render params template")
         exit(1)
     else:
-        args.cluster_id_map = os.path.abspath(args.cluster_id_map)
+        args.input_file = os.path.abspath(args.input_file)
```
rbdavid: This should move to before line 39? Why is this happening as a part of an else statement checking for failure or success?

pipelines/gnd/gnd.nf
```diff
@@ -1,22 +1,140 @@
...
+        path "init_blast.out"
+
+    script:
+    def blast_num_matches_arg = params.import_blast_num_matches ? "-b ${params.import_blast_num_matches}" : ""
+    def blast_evalue_arg = params.import_blast_evalue ? "-e ${params.import_blast_evalue}" : ""
+
+    """
+    blastall -p blastp -i ${sequence_file} -d ${params.import_blast_fasta_db} -m 8 ${blast_evalue_arg} ${blast_num_matches_arg} -o init_blast.out
+    if [[ ! -s init_blast.out ]]; then
+        echo "BLAST did not return any matches.  Verify that the sequence is a protein and not a nucleotide sequence."
```
rbdavid: Can this be echo'd to stderr? I guess just by adding `>&2` to the end.

pipelines/gnd/gnd.nf
```diff
@@ -1,22 +1,140 @@
...
+    input:
+        path sequence_file
+        path blast_file
+        val seq_ver
+    output:
+        path "accession_ids.tab", emit: "source_ids"
+        path "sequence_meta.tab", emit: "source_meta"
+
+    script:
+    def common_args = "--efi-config ${params.efi_config} --efi-db ${params.efi_db} --mode blast --sequence-version ${seq_ver}"
```
rbdavid: I guess I don't really understand the use of this `common_args` instead of just explicitly calling these arguments within the `perl ${projectDir}/../ ...` lines below. Or add the `--source-meta-file` and `--source-ids-file` lines to the `common_args` call?

pipelines/gnd/gnd.nf
```diff
@@ -1,22 +1,140 @@
...
+process parse_fasta_for_ids {
+    input:
+        path fasta_file
+        val seq_ver
+    output:
+        path "accession_ids.tab", emit: "source_ids"
+        path "sequence_meta.tab", emit: "source_meta"
+
+    script:
+    def common_args = "--efi-config ${params.efi_config} --efi-db ${params.efi_db} --mode fasta --sequence-version ${seq_ver}"
```
rbdavid: See above. There's no special logic being set up by the `common_args` variable so why not just include these arguments in the script block below?

pipelines/gnd/gnd.nf
```diff
@@ -1,22 +1,140 @@
...
+process parse_accession_for_ids {
+    input:
+        path accession_file
+        val seq_ver
+    output:
+        path "accession_ids.tab", emit: "source_ids"
+        path "sequence_meta.tab", emit: "source_meta"
+
+    script:
+    def common_args = "--efi-config ${params.efi_config} --efi-db ${params.efi_db} --mode accessions --sequence-version ${seq_ver}"
```
rbdavid: See above.

pipelines/gnd/gnd.nf
```diff
@@ -1,22 +1,140 @@
...
+
+    if (params.import_mode == "accessions") {
+        parse_data = parse_accession_for_ids(input_file_ch, seq_ver)
+    } else if (params.import_mode == "blast") {
+        blast_results = run_blast(input_file_ch)
+        parse_data = parse_blast_results_for_ids(input_file_ch, blast_results, seq_ver)
+    } else if (params.import_mode == "fasta") {
+        parse_data = parse_fasta_for_ids(input_file_ch, seq_ver)
+    } else {
+        error "Mode '${params.import_mode}' is invalid"
```
rbdavid, line 136: I have tried to find nextflow documentation on the `error` syntax/command. If possible, being able to throw an explicit return code along with the error message might be very helpful. I don't know if that's possible with the nf standard `error` call you are using here.

rbdavid (approved):

bin/create_gnd_nextflow_params.py
nilsoberg: This line is ensuring that the parameter passed via `params.json` is an absolute path.  Whether or not this line occurs before the check for existence or after doesn't affect the validation code, since the parameter passed to the script should exist already.  I moved it to an `else` block.

pipelines/gnd/gnd.nf
nilsoberg: That code occurs in groovy, allowing for a cleaner verification that `params.sequence_version` exists (rather than using some Bash logic).  Or am I misunderstanding?

nilsoberg: Updated the logic to make intent clearer

pipelines/gnd/gnd.nf
nilsoberg: It looks like `error` stops the pipeline and returns a non-zero exit code to the shell.  We could write a custom error handler, but this would be useful across all the EST pipelines.  Something like:

```
workflow.onComplete {
    println "Pipeline completed at: $workflow.complete"
    println "Execution status: ${ workflow.success ? 'OK' : 'failed' }"
    println "Exit status: ${ workflow.exitStatus }"
}
```

But this is out of the scope of this PR and I would prefer to leave as is, especially since import mode is set programmatically in most usages.

rbdavid: I agree that it is out of scope for this PR. 

I was not imagining that the error/exit status would be reported in the `.onComplete` call for the workflow. If I understand correctly, calling `error` is the clean way to end the workflow abruptly. Maybe the `error` message can contain a subprocess-specific exit code in the string. Then a failed run of the workflow can be detected and the subsequent error string can be parsed to ID the failure and narrow down the scope for debugging/fixing that specific run.

pipelines/clusteranalysis/clusteranalysis.nf
rbdavid, line 1: Why did this get added to this PR? This is sufficiently large/important to be its own PR, right?

rbdavid: I have not reviewed this code.

pipelines/gnd/gnd.nf
```diff
@@ -12,9 +12,9 @@ process create_gnd {
     script:
     """
     perl ${projectDir}/create_gnd.pl \
-        --config ${params.efi_config} \
-        --db-name ${params.efi_db} \
-        --cluster-map ${id_list} \
+        --config "${params.efi_config}" \
+        --db-name "${params.efi_db}" \
+        --cluster-map "${id_list}" \
         --nb-size ${params.nb_size} \
```
rbdavid, line 18: Why was this line's variable's double quoting skipped? See my below comment for more intent explained.

pipelines/gnd/gnd.nf
```diff
@@ -34,9 +34,9 @@ process run_blast {
     def blast_evalue_arg = params.import_blast_evalue ? "-e ${params.import_blast_evalue}" : ""
```
rbdavid: Just to clarify, the `conf/default_params.config` file contains default values for these arguments, so these def lines will never assign an empty string to the variable. There's not much use in these `def` lines with this specific logic.

pipelines/gnd/gnd.nf
```diff
@@ -34,9 +34,9 @@ process run_blast {
...
 
     """
-    blastall -p blastp -i ${sequence_file} -d ${params.import_blast_fasta_db} -m 8 ${blast_evalue_arg} ${blast_num_matches_arg} -o init_blast.out
+    blastall -p blastp -i "${sequence_file}" -d "${params.import_blast_fasta_db}" -m 8 ${blast_evalue_arg} ${blast_num_matches_arg} -o init_blast.out
```
rbdavid: So, I'm a bit confused why all variables aren't being wrapped with double quotes. There's no validation (on the EST backend code side of things) to check that parameters defined in the `params.json` or in the config files are the correct typing, so surrounding their usage in bash scripts with double quotes would defend against unwanted expansions, redirections, etc.

pipelines/gnd/gnd.nf
rbdavid: I'm not sure what Groovy has to do with this. Moving the definition of input arguments, albeit only five lines above, seems to only obfuscate what's being called in the bash script's lines. Idk, its not a huge deal, it just seems unnecessary.

pipelines/gnd/gnd.nf
rbdavid: Now, if we want to remove these params from having default values that are set in the config file, that's a conversation. I don't think we should do that, but its a conversation we could have.

pipelines/gnd/gnd.nf
```diff
@@ -52,12 +52,14 @@ process parse_blast_results_for_ids {
         path "sequence_meta.tab", emit: "source_meta"
 
     script:
-    def common_args = "--efi-config ${params.efi_config} --efi-db ${params.efi_db} --mode blast --sequence-version ${seq_ver}"
+    def seq_ver_arg = params.sequence_version ? "--sequence-version ${params.sequence_version}" : ""
```
rbdavid: See above about default value already being set. This logic will never be used, currently.

rbdavid (commented):
There's clusteranalysis.nf code in here that should not be. Can we remove that and push to a separate PR?

pipelines/gnd/gnd.nf
nilsoberg: Could you give an example of what you had on mind regarding moving of input arguments on line 55?

pipelines/clusteranalysis/clusteranalysis.nf
nilsoberg: This CA code should never have been committed. I'm not sure how it happened but that will be addressed.

pipelines/gnd/gnd.nf
nilsoberg: In retrospect, I think quoting of most of those file names is redundant because they are internal names passed between nextflow processes. Only input file paths or paths that come from the params file should be passed as quoted arguments. If an argument won't contain a space, e.g. neighborhood size, then quoting it is not necessary. I'm going to remove most of these quotation marks in keeping with the rest of the code.

rbdavid: There's nothing stopping someone from settings `params.nb_size = "hello world &&"` (or something equivalent; that example code isn't following json formatting, of course) in the params.json, which would then be pasted into this bash line and cause errors.

pipelines/gnd/gnd.nf
nilsoberg: I think it's possible for a user of the command line tools to provide a malformed or incomplete config file, and that's the intent behind this code. I believe the original parameters had a different name and it was possible for them to not be present.

pipelines/gnd/gnd.nf
nilsoberg: Will sequence_version will always be set?

rbdavid: I don't think any of these `def` lines that you've added are needed. For example, the `params.sequence_version` has a default value set in the `conf/default_params.config` file, so the empty string result of this ternary logic is never used. So, the def line is just not needed.

rbdavid: > Will sequence_version will always be set?

yes. in `conf/default_params.config`

nilsoberg: I was not aware that all parameters were present in the file. Then yes, this logic is redundant.

pipelines/gnd/gnd.nf
rbdavid: The purpose of the default_params.config file is to hard-set the defaults, enabling implementation/installation specific control of the various parameters that can reasonably have a default value. If those are malformed in anyway, that's an implementation error not a run error.

pipelines/gnd/gnd.nf
nilsoberg: Input to the Web forms is sanitized so no invalid inputs (like your example) will come from website user input. If a user is creating their own parameters file on an HPC and attempting to do malicious things, then there are bigger problems than sanitizing all arguments. If the backend is not a trusted service, then this as an issue we should be concerned about. As it stands, I don't think it's an issue.

rbdavid: We should not be developing the code assuming sanitized parameters from a different code base are going to be receieved. I think using double quotes around parameter calls that are used directly in bash script sections of the workflow is the lowest/easiest defense against any sort of malicious or innocent mistakes. Why wouldn't we implement them?

nilsoberg:
Good point on escaping command line arguments.  I added quotation marks around parameters that are potentially paths.  Merging to stay on track.

pipelines/gnd/gnd.nf
```diff
@@ -12,13 +11,13 @@ process create_gnd {
...
-        --db-name "${params.efi_db}" \
-        --cluster-map "${id_list}" \
+        --efi-config "${params.efi_config}" \
+        --efi-db "${params.efi_db}" \
+        --sequence-version ${params.sequence_version} \
+        --cluster-map ${id_list} \
         --nb-size ${params.nb_size} \
         --gnd gnd.sqlite \
-        --stats stats.json \
-        --sequence-version ${seq_ver}
```
rbdavid, line 21: Why was this line deleted? It needs to still be an argument to the `create_gnd.pl` script, correct? What it needed to be changed to is `--sequence-version "${params.sequence_version}"`

pipelines/gnd/gnd.nf
```diff
@@ -3,7 +3,6 @@ process create_gnd {
     publishDir params.final_output_dir, mode: 'copy'
     input:
         path id_list 
-        val seq_ver
```
rbdavid, line 6: Its fine to delete this line.

rbdavid (commented):
My review below was a quick pass and I missed the replacement line for calling the `--sequence-version` cli argument, which was moved a few lines up. I spoke too quickly.

pipelines/gnd/gnd.nf
nilsoberg: Look at the latest code

Issue #259, opened 2026-02-11
Implement duckdb memory allocation parameters in render_prereduce_sql_template.py and sql template


rbdavid:
If I understand correctly, the below line does not include any usage of `--duckdb-memory-limit ${params.duckdb_memory_limit} --duckdb-temp-dir /scratch/duckdb-${params.job_id}` because the following `duckdb` call is only processing a single .tab.parquet file from one all_by_all_blast subprocess. But, memory allocation should be explicitly controlled and the temp-directory should be defined to avoid cases where a shard's result file overflows the available duckdb memory and/or duckdb writes temp files in an incorrect storage space. Available memory and temp space is implementation specific, so what is left undefined may result in poorly performing or unexplained behavior on the different systems that the code is being installed on. 

https://github.com/EnzymeFunctionInitiative/EST/blob/42faa7fe964635918848115c7fcd818c8c065748/pipelines/est/subworkflows/all_by_all.nf#L30

rbdavid:
Closing since this was implemented in #260

PR #260, opened 2026-02-11
258/9 - Keep self-alignments and add duckdb allocation limits

### LINKED ISSUE #258: blastreduce code should not remove self-alignment lines
The self-alignment lines in the all-by-all blast result files are important to maintain, specifically for when the sequence set is compressed/reduced before running the alignments. The below line needs to be removed: 

https://github.com/EnzymeFunctionInitiative/EST/blob/2c470f520844460b5e8982bfc4e41b33651cc09b/pipelines/est/templates/reduce-template.sql#L10-L10

In the same PR, the logic of the blastreduce's reduce-template.sql code should be verified to be replicating the original perl code's results. Comments in that sql template file says that it does not exactly match the original code. 

https://github.com/EnzymeFunctionInitiative/EST/blob/2c470f520844460b5e8982bfc4e41b33651cc09b/pipelines/est/templates/reduce-template.sql#L14-L17

I haven't done any sort of speed testing but I highly doubt processing blast result file(s) is the bottle neck for the EST workflow. Is this duckdb strategy worth the slight inaccuracies?

rbdavid:
closes #258 and #259 

For #258, removed the guilty lines in the sql template files that skip over lines in the `.tab.parquet` and `1.out.parquet` files where `qseqid = sseqid`. Those are important to keep, whether the initial sequence set was reduced to representative sequences or not. 

For #259, the `prereduce-template.sql` file did not have template lines to enable us to specify a memory limit and a temp dir where duckdb would write temp files. Apparently, when a memory limit is not set, duckdb uses 80% of the physical RAM of the system; when duckdb is running on a shared compute resource, this is very likely to be waaaaayyyyy over the allocated RAM amount.


rbdavid:
Here's the blog post that explains duckdb's memory management strategy: https://duckdb.org/2024/07/09/memory-management

Here's the duckdb documentation where the default memory limit is set: https://duckdb.org/docs/stable/operations_manual/limits

rbdavid:
There will be some slight clashes with the #255 PR due to the few lines in `all_by_all.nf` that have been changed between those two branches. Those should be easy to resolve.

pipelines/est/subworkflows/all_by_all.nf
```diff
@@ -27,7 +27,7 @@ process all_by_all_blast {
     python $projectDir/axa_blast/transcode_blast.py --blast-output ${frac}.tab
 
     # in each row, ensure that qseqid < sseqid lexicographically
-    python $projectDir/axa_blast/render_prereduce_sql_template.py --blast-output ${frac}.tab.parquet --sql-template $projectDir/templates/prereduce-template.sql --output-file ${frac}.tab.sorted.parquet --duckdb-temp-dir /scratch/duckdb-${params.job_id} --sql-output-file prereduce.sql
+    python $projectDir/axa_blast/render_prereduce_sql_template.py --blast-output ${frac}.tab.parquet --sql-template $projectDir/templates/prereduce-template.sql --output-file ${frac}.tab.sorted.parquet --duckdb-memory-limit "${params.duckdb_memory_limit}" --duckdb-temp-dir /scratch/duckdb-"${params.job_id}"  --sql-output-file prereduce.sql
```
nilsoberg: With the change to the way the --duckdb-temp-dir argument is being quoted, I'm pretty sure that the change `--duckdb-temp-dir /scratch/duckdb-"${params.job_id}"` is going to produce files like `/scratch/duckdb-"33"` instead of files `/scratch/duckdb-33`.

nilsoberg: It's possible that there is an issue with temp dir being shared between multiple processes.  What should be used instead of `params.job_id` is `task.hash`.  This also doesn't need to be escaped.

pipelines/est/subworkflows/all_by_all.nf
```diff
@@ -76,7 +76,7 @@ process demultiplex {
     output:
         path '1.out.parquet'
     """
-    echo "COPY (SELECT * FROM read_parquet('$blast_parquet')) TO 'mux.out' (FORMAT CSV, DELIMITER '\t', HEADER false);" | duckdb
+    echo "SET memory_limit = '"${params.duckdb_memory_limit}"'; SET temp_directory = '/scratch/duckdb-"${params.job_id}"'; SET threads TO 1; COPY (SELECT * FROM read_parquet('$blast_parquet')) TO 'mux.out' (FORMAT CSV, DELIMITER '\t', HEADER false);" | duckdb
```
nilsoberg: Similar quoting issue as before.  Validation of this parameter needs to occur in a separate place since there isn't robust SQL parameterization in this environment.

nilsoberg:
```suggestion
    echo "SET memory_limit = '${params.duckdb_memory_limit}'; SET temp_directory = '/scratch/duckdb-${task.hash}'; SET threads TO 1; COPY (SELECT * FROM read_parquet('$blast_parquet')) TO 'mux.out' (FORMAT CSV, DELIMITER '\t', HEADER false);" | duckdb
```

pipelines/est/subworkflows/reporting.nf
```diff
@@ -16,7 +16,7 @@ process compute_stats {
     python $projectDir/statistics/conv_ratio.py --blast-output $blast_parquet --fasta $fasta_file --output conv_ratio.json
 
     # compute boxplot stats and evalue.tab
-    python $projectDir/statistics/render_boxplotstats_sql_template.py --blast-output $blast_parquet --duckdb-temp-dir /scratch/duckdb-${params.job_id} --boxplot-stats-output boxplot_stats.parquet --evalue-output evalue.tab --sql-template $projectDir/templates/boxplotstats-template.sql --sql-output-file boxplotstats.sql
+    python $projectDir/statistics/render_boxplotstats_sql_template.py --blast-output $blast_parquet --duckdb-memory-limit "${params.duckdb_memory_limit}" --duckdb-temp-dir /scratch/duckdb-"${params.job_id}" --boxplot-stats-output boxplot_stats.parquet --evalue-output evalue.tab --sql-template $projectDir/templates/boxplotstats-template.sql --sql-output-file boxplotstats.sql
```
nilsoberg: Same quoting issue as in previous comments.

nilsoberg:
```suggestion
    python $projectDir/statistics/render_boxplotstats_sql_template.py --blast-output $blast_parquet --duckdb-memory-limit "${params.duckdb_memory_limit}" --duckdb-temp-dir /scratch/duckdb-${task.hash} --boxplot-stats-output boxplot_stats.parquet --evalue-output evalue.tab --sql-template $projectDir/templates/boxplotstats-template.sql --sql-output-file boxplotstats.sql
```

pipelines/est/subworkflows/all_by_all.nf
nilsoberg:
```suggestion
    python $projectDir/axa_blast/render_prereduce_sql_template.py --blast-output ${frac}.tab.parquet --sql-template $projectDir/templates/prereduce-template.sql --output-file ${frac}.tab.sorted.parquet --duckdb-memory-limit "${params.duckdb_memory_limit}" --duckdb-temp-dir /scratch/duckdb-${task.hash}  --sql-output-file prereduce.sql
```

pipelines/est/templates/prereduce-template.sql
```diff
@@ -1,11 +1,12 @@
+SET memory_limit = '$mem_limit';
+SET temp_directory = '$duckdb_temp_dir';
 SET threads TO 1;
```
nilsoberg, line 3: Set threads to `$threads`?

nilsoberg (commented):
It was important to add these arguments to duckdb.  
There are some issues with escaping, and I suggested changes.
Could you document with concrete results that these changes fixed the potential issue?

pipelines/est/subworkflows/all_by_all.nf
rbdavid: That's a great idea. I've committed the suggestion you made below. Will get this updates as well.

pipelines/est/templates/prereduce-template.sql
rbdavid: `$threads` is not a template string that maps to any input arguments for the `render_..._sql_tempalte.py` scripts. Generally, duckdb has always been run using a single thread in the EST code. If different threading allocations need to be developed for, that might constitute a new feature that gets its own issue/PR. Let's chat about it!

pipelines/est/subworkflows/all_by_all.nf
rbdavid: The `script` [section](https://www.nextflow.io/docs/latest/process.html#script) block in the process is using the bash shell, so the command actually run by the lines between the ` ``` ` lines should match exactly what `echo` of those lines would print. Below is an example of the line, run from the cli so I'm ignoring `params.` variable prefixes.

``` bash
$ frac=42; duckdb_memory_limit="4GB"; job_id=50; duckdb_temp_dir="scratch/duckdb" 
$ echo "python $projectDir/axa_blast/render_prereduce_sql_template.py --blast-output ${frac}.tab.parquet --sql-template $projectDir/templates/prereduce-template.sql --output-file ${frac}.tab.sorted.parquet --duckdb-memory-limit "${duckdb_memory_limit}" --duckdb-temp-dir "${duckdb_temp_dir}"-"${job_id}"  --sql-output-file prereduce.sql"

> python /axa_blast/render_prereduce_sql_template.py --blast-output 42.tab.parquet --sql-template /templates/prereduce-template.sql --output-file 42.tab.sorted.parquet --duckdb-memory-limit 4GB --duckdb-temp-dir /scartch/duckdb-50  --sql-output-file prereduce.sql

```

The only way we get, for example, `/scratch/duckdb-"33"` is if the quotation marks are escaped.

pipelines/est/subworkflows/all_by_all.nf
rbdavid: As above, this is just a bash `script` block. The quotation marks are doing as designed and printing the variable's literal value, while preventing potential wildcard expansions and other undesired behaviors.

pipelines/est/subworkflows/all_by_all.nf
nilsoberg: The reason you're seeing the output you see is because you are using double quotation marks.  If you do `echo '...'` instead you will see the problem.  Bash `echo` joins things together that are adjacent (e.g. `echo "A"B"C"` === `echo "ABC"`).

nilsoberg (approved):

pipelines/est/subworkflows/all_by_all.nf
rbdavid: Running below as a bash script:
```
set -x

frac=42; duckdb_memory_limit="4GB"; job_id=50; duckdb_temp_dir="scratch/duckdb"; projectDir=~/EST/pipelines/est
python $projectDir/axa_blast/render_prereduce_sql_template.py --blast-output ${frac}.tab.parquet --sql-template $projectDir/templates/prereduce-template.sql --output-file ${frac}.tab.sorted.parquet --duckdb-memory-limit "${duckdb_memory_limit}" --duckdb-temp-dir "${duckdb_temp_dir}"-"${job_id}"  --sql-output-file prereduce.sql

```
Returns:
```
+ frac=42
+ duckdb_memory_limit=4GB
+ job_id=50
+ duckdb_temp_dir=scratch/duckdb
+ python ~/EST/pipelines/est/axa_blast/render_prereduce_sql_template.py --blast-output 42.tab.parquet --sql-template ~/EST/pipelines/est/templates/prereduce-template.sql --output-file 42.tab.sorted.parquet --duckdb-memory-limit 4GB --duckdb-temp-dir scratch/duckdb-50 --sql-output-file prereduce.sql
```

So, I'm not exactly sure what the issue is with the double quotes. I'll remove them just so we aren't continuing this conversation.

PR #261, opened 2026-02-11
33 cluster analysis WIP refactor histograms

nilsoberg:
Changes in the PR allow use of length histogram generation code by both cluster analysis and est pipelines.  The first commit in the list, 2ceca4f, regarding `clusteranalysis.nf`, is not relevant so it should not be reviewed now (it has been significantly changed in the main 33-cluster-analysis-workflow branch and will come as part of a separate PR).

Changes made:

* The entire title can be provided as an argument to the `plot_length_data.py` script, avoiding the requirement for a job ID (needed for cluster analysis plot generation)
* It is no longer required to pass an accession ID table file to `compute_length_histogram.py`. If this is not provided, then the length histogram is computed based on the FASTA file alone. This required a significant rewrite of the file.


nilsoberg:
Closing to change base branch

PR #265, opened 2026-02-13
Update Nextflow default_params.config file for Cluster Analysis pipeline

nilsoberg:
- Settings relating to MUSCLE and output options
- Update params creation script for default MSA min/max cluster size

Up for discussion is default number of CPUs for MUSCLE.  See comments.

### DIFF
--- a/conf/docker.config
+++ b/conf/docker.config
@@ -23,6 +23,10 @@ process {
     withName: 'filter_blast|create_full_ssn' {
         memory = 16.GB
     }
+
+    withName: muscle_align {
+        cpus = 8
+    }
 }
 
 /*
--- a/conf/docker.config
+++ b/conf/docker.config
@@ -25,7 +25,7 @@ process {
     }
 
     withName: muscle_align {
-        cpus = 8
+        cpus = 1
     }
 }
 
--- a/bin/create_clusteranalysis_nextflow_params.py
+++ b/bin/create_clusteranalysis_nextflow_params.py
@@ -8,6 +8,8 @@ import os
 import shared_args
 
 NXF_SCRIPT = "pipelines/clusteranalysis/clusteranalysis.nf"
+DEFAULT_MAX_MSA_SEQ = 750
+DEFAULT_MIN_MSA_SEQ = 3
 
 def add_args(parser: argparse.ArgumentParser):
     """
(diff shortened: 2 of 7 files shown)


nilsoberg:
There are several issues with MUSCLE in this tool.  The first issue is the inconsistency between versions MUSCLE v3 and v5 (v5 PPP and v5 Super5).  A simple short test showed wildly inconsistent alignments between the three cases.  The second issue is performance; the tests here were run using 750 sequences on a slow PC but show the differences between the modes.  The legacy EFI tools used MUSCLE v3 which gave decent performance (7:35) with one thread.  v5 PPP took 52:16 with 8 threads.  v5 Super5 (should be used for larger alignments) took 2:48 with 8 threads.  The third issue is deployment; MUSCLE v3 is not on GitHub and is subject to the reliability of the hosting server (www.drive5.com); the binary would need to be included in the repository in `vendor/bin` or in another GitHub repository specifically for dependencies.  MUSCLE v5 is available on GitHub, both source code and Linux binary making it easy to deploy.

I updated `conf/docker.config` with a section for MUSCLE to specify how many threads it is allowed to use.  Due to the discrepancy in results, I am not certain we should go with MUSCLE v5, so there is a MUSCLE version flag in the default params file to switch between the two.  The default is version 3.

Further examination of results is needed, but is out of the scope of this initial release (Feb-March).  See issue #266.

rbdavid:
> MUSCLE v3 is not on GitHub and is subject to the reliability of the hosting server ([www.drive5.com](http://www.drive5.com)); the binary would need to be included in the repository in `vendor/bin` or in another GitHub repository specifically for dependencies. 

Couldn't we host the source code tar ball or binary file (whatever is currently stored in `vendor/bin`) on our downloads rsync/ftp site. That avoids hosting the code on github, where upkeep might be expected (unreasonably). And would be easily implemented as a `wget` call in the Dockerfile when deploying the code.

conf/docker.config
rbdavid, line 1: This change should be implemented to the other config files, right? Using the correct executors and what not.

rbdavid (approved):

conf/docker.config
nilsoberg: Yes. Done

nilsoberg:
> > MUSCLE v3 is not on GitHub and is subject to the reliability of the hosting server ([www.drive5.com](http://www.drive5.com)); the binary would need to be included in the repository in `vendor/bin` or in another GitHub repository specifically for dependencies.
> 
> Couldn't we host the source code tar ball or binary file (whatever is currently stored in `vendor/bin`) on our downloads rsync/ftp site. That avoids hosting the code on github, where upkeep might be expected (unreasonably). And would be easily implemented as a `wget` call in the Dockerfile when deploying the code.

MUSCLE could be put in a GitHub repo.

PR #272, opened 2026-02-17
33 cluster analysis

nilsoberg:


conf/docker.config
```diff
@@ -23,6 +23,10 @@ process {
     withName: 'filter_blast|create_full_ssn' {
         memory = 16.GB
     }
+
+    withName: muscle_align {
+        cpus = 8
+    }
```
rbdavid, line 29: Other config files should also have this block. And for PBSPro/Slurm/etc configs, those blocks will need to call the specific executor since the default parent job allocation likely won't allocate 8 cpus.

Dockerfile
```diff
@@ -1,31 +1,81 @@
...
 COPY lib/pyEFI /lib/pyEFI
 
+# set up python environment
+RUN pip3 install -r app/requirements.txt
+
+# set up Perl environment
+RUN cpanm --installdeps .
+
 # install blastall
 RUN curl -o /opt/blast-2.2.26.tar.gz https://ftp.ncbi.nlm.nih.gov/blast/executables/legacy.NOTSUPPORTED/2.2.26/blast-2.2.26-x64-linux.tar.gz; \
```
rbdavid: As below in the new code, should the semi-colons be replaced by `&&` so that the lines are only performed if the previous line was successful?

Dockerfile
```diff
@@ -1,31 +1,81 @@
...
     unzip /opt/duckdb/duckdb.zip -d /opt/duckdb/; \
     rm /opt/duckdb/duckdb.zip
-ENV PATH="${PATH}:/opt/duckdb/"
 
 # install CD-HIT
 RUN curl -L -o /opt/cd-hit.tar.gz https://github.com/weizhongli/cdhit/releases/download/V4.8.1/cd-hit-v4.8.1-2019-0228.tar.gz; \
     tar xzf /opt/cd-hit.tar.gz -C /opt; \
     rm /opt/cd-hit.tar.gz; \
     cd /opt/cd-hit-v4.8.1-2019-0228; make; mkdir bin; mv cd-hit bin;
-ENV PATH="${PATH}:/opt/cd-hit-v4.8.1-2019-0228/bin"
```
rbdavid, line 25: I think the only logic to keep these lines here is that they localize the cd-hit addition to the Path to the block where cd-hit is installed. If this type of line is removed and replaced with a single `ENV PATH=` line at the end of the file, we make changes up here also require us to remember to make changes down there.

Dockerfile
```diff
@@ -1,31 +1,81 @@
...
+RUN mkdir -p /opt/seqkit && \
+    curl -L -o /opt/seqkit/seqkit.tar.gz https://github.com/shenwei356/seqkit/releases/download/${SEQKIT_VER}/seqkit_linux_amd64.tar.gz && \
+    tar -zxvf /opt/seqkit/seqkit.tar.gz -C /opt/seqkit && \
+    rm /opt/seqkit/seqkit.tar.gz
+
+# install MUSCLE (v5.1)
+RUN mkdir -p /opt/muscle && \
+    curl -L -o /opt/muscle/muscle5 https://github.com/rcedgar/muscle/releases/download/v5.3/muscle-linux-x86.v5.3 && \
+    chmod +x /opt/muscle/muscle5
+COPY vendor/bin/muscle-3.8.31 /opt/muscle/muscle3
```
rbdavid: I worry a bit about stashing this binary file inside of this repository. Instead, can we host it on our rsync/ftp downloads page and curl it here instead?

vendor/bin/muscle-3.8.31
rbdavid, line 1: I don't think this should be stashed in the EST repository.

pipelines/shared/python/compute_length_histogram.py
```diff
@@ -24,56 +24,59 @@ def parse_fasta(file_handle: TextIO) -> Iterator[Tuple[str, str]]:
...
-                
-                # For 'uniprot' type, we consider all IDs in the first column.
-                if seq_type == 'uniprot':
-                    ids.add(uniprot_id)
-                # For 'uniref' types, we only add the UniProt ID if the corresponding
-                # UniRef column is not empty.
-                elif parts[target_col]:
-                    ids.add(uniprot_id)
+                parts = line.split('\t')
+                if len(parts) > target_col:
```
rbdavid, line 62: Is there the possibility for line(s) not having the same number of tab-separated columns as the rest of the file? Or is the number of columns hard-set when the file(s) are written?

Like, can an accession_table.tab file contain information for an accession ID that does not have a UniRef90/50 id(s)?

pipelines/shared/python/compute_length_histogram.py
```diff
@@ -24,56 +24,59 @@ def parse_fasta(file_handle: TextIO) -> Iterator[Tuple[str, str]]:
...
+                    val = parts[target_col].strip()
+                    if val:
+                        ids.add(val)
```
rbdavid, line 65: This is a relatively slow, line-by-line processing. It works. But we could use any dataframe-style processing library to read the csv and processes the whole file in two lines, using optimized C code instead of raw python loops, etc.

```
import pandas
df = pandas.read_csv(input_file, sep='\t', header=0 if has_header_line else None).fillna("Missing")
ids = set(df.iloc[:,target_col])

# remove any instance where the column was an empty and was replaced by "Missing"
if "Missing" in ids:
    ids.remove("Missing")
```

pipelines/shared/python/compute_length_histogram.py
rbdavid: If this bool check returns False, shouldn't we raise an error, indicating a malformed file?

pipelines/shared/python/compute_length_histogram.py
rbdavid: Also, if the `target_col` value is incorrectly set, the `ids = set(df.iloc ...` line will appropriately raise an exception, unlike the current code which will just return an empty set and pass it on to be written to file.

pipelines/shared/python/compute_length_histogram.py
```diff
@@ -24,56 +24,59 @@ def parse_fasta(file_handle: TextIO) -> Iterator[Tuple[str, str]]:
     if header:
         yield (header, ''.join(sequence))
 
-def parse_accession_table(accession_table: str, seq_type: str) -> Set[str]:
+def parse_file(input_file: str, target_col: int, has_header_line: bool = True) -> Set[str]:
     """
-    Parses a three-column accession table file and gets the list of sequences that correspond
-    to the given sequence type (e.g. uniprot, uniref90, uniref50).  The table file has a header
-    column.
+    Parses a file and extracts IDs from the given column index. The file can have an optional header column.
```
rbdavid: Doc strings should be fixed to better match the pyEFI standards.

pipelines/shared/python/compute_length_histogram.py
```diff
@@ -24,56 +24,59 @@ def parse_fasta(file_handle: TextIO) -> Iterator[Tuple[str, str]]:
...
+        print(f"Error: input ID file not found at '{input_file}'", file=sys.stderr)
         sys.exit(1)
 
     return ids
 
-def compute_sequence_lengths(fasta_file: str, valid_ids: Set[str]) -> List[int]:
+def parse_accession_table(accession_table: str, seq_type: str) -> Set[str]:
     """
-    Process the FASTA file and collect lengths
+    Parses a standard three-column accession table file output by the 'est' pipeline, and gets
```
rbdavid: Doc strings should be fixed to better match the pyEFI standards.

pipelines/shared/python/compute_length_histogram.py
```diff
@@ -24,56 +24,59 @@ def parse_fasta(file_handle: TextIO) -> Iterator[Tuple[str, str]]:
...
+    col_map = {'uniprot': 0, 'uniref90': 1, 'uniref50': 2}
+    target_col = col_map[seq_type]
+
+    return parse_file(accession_table, target_col, True)
+
+def compute_sequence_lengths(fasta_file: str, valid_ids: Set[str] = None) -> List[int]:
+    """
+    Process the FASTA file and collect lengths. If 'valid_ids' is a Set, then only sequences
+    with IDs in the Set are included in the output. If 'valid_ids' is None, then all
+    sequences are included.
```
rbdavid: This code is somewhat redundant with `pyEFI.transcode.fasta_to_parquet()` function (https://github.com/EnzymeFunctionInitiative/EST/blob/bd0aa67019b645bc4026f52256c470cb32e186ed/lib/pyEFI/pyEFI/transcode.py#L51), which uses biopython (`from Bio import SeqIO`) to process fasta files cleanly. I think using biopython as the standard parser for fasta files is safer than this and the `parse_fasta` function defined above.

pipelines/shared/python/compute_length_histogram.py
```diff
@@ -85,20 +88,25 @@ def compute_sequence_lengths(fasta_file: str, valid_ids: Set[str]) -> List[int]:
...
+    if accession_table and seq_type:
+        valid_ids = parse_accession_table(accession_table, seq_type)
+    else:
+        valid_ids = None
 
     sequence_lengths = compute_sequence_lengths(fasta_file, valid_ids)
 
     if not sequence_lengths:
-        print(f"Warning: No sequences were selected for histogram generation for type '{seq_type}'. Output will be empty.", file=sys.stderr)
+        print(f"Warning: No sequences were selected for histogram generation{seq_type_str}. Output will be empty.", file=sys.stderr)
```
rbdavid, line 167: This should raise an exception, correct? Otherwise, an empty histogram file is being written here and the buck is passed on to the plot_length_histogram script where an error/exception is thrown because there is no data to plot.

rbdavid: That plot_length_histogram is called in a separate nextflow process block, making the debugging of this issue even more difficult than it should be.

bin/create_clusteranalysis_nextflow_params.py
```diff
@@ -0,0 +1,127 @@
...
+                if any(x < 0 or x > 100 for x in pid_list):
+                    print(f"--pid-threshold values must be >= 0 and <= 100")
+                    fail = True
+                args.pid_thresholds = pid_list
+            except ValueError:
+                print(f"--pid-threshold must be a comma-separated list of percentages from 0-100")
+                fail = True
+
+        if args.residues:
+            r_list = args.residues.split(',')
```
rbdavid: Should elements in this list also have white space stripped? As done on 62 for the pid_list.

rbdavid (commented):
Just some suggestions.

conf/docker.config
nilsoberg: I added this to all config files.  I set the default value to `1` since we're using MUSCLE v3 by default.

Dockerfile
nilsoberg: Yes

Dockerfile
nilsoberg: It's now at https://github.com/EnzymeFunctionInitiative/MUSCLE/releases/tag/v3.8.1551-drive5

vendor/bin/muscle-3.8.31
nilsoberg: It's now at https://github.com/EnzymeFunctionInitiative/MUSCLE/releases/tag/v3.8.1551-drive5

pipelines/shared/python/compute_length_histogram.py
nilsoberg: It's possible.  No exception is necessary, because some times a UniProt entry may not have a UniRef entry.  In that case the sequence is skipped.

pipelines/clusteranalysis/logo/convert_hmm_to_skylign_json.py
```diff
@@ -0,0 +1,112 @@
...
+    hmm_symbols = hmm.alphabet.symbols
+    
+    # 3. Extract Probabilities
+    # hmm.match_emission is a matrix of negative natural log probabilities (scores)
+    # Shape: (Length + 1, Alphabet_Size)
+    # We convert them back to probabilities: p = exp(-score)
+    mat_emissions = np.asarray(hmm.match_emissions)
+    probs_matrix = np.exp(-mat_emissions)
+
+    # FIX 2: Check if name is bytes before decoding (Newer pyhmmer returns str)
```
rbdavid: What does "FIX 2" mean in this case?

pipelines/clusteranalysis/logo/convert_hmm_to_skylign_json.py
```diff
@@ -0,0 +1,112 @@
+
+import argparse
+import json
+import math
+import numpy as np
+import pyhmmer
+
+def convert_hmm_to_skylign_json(hmm_path, output_json_path):
+    """
+    Parses an HMM, calculates Information Content (Skylign style),
```
rbdavid: Doc strings

pipelines/clusteranalysis/logo/plot_hmm_weblogo.py
```diff
@@ -0,0 +1,62 @@
+import pyhmmer
+import logomaker
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+
+def plot_hmm_logo(hmm_path, output_file):
```
rbdavid: No doc strings at all

pipelines/clusteranalysis/logo/plot_hmm_weblogo.py
rbdavid, line 1: This script and the `convert_hmm_to_skylign_json.py` script share a ton of code. Can we concentrate the shared code and isolate the plotting (in this script) and the json-writing (in the other) instead of rewriting the same math in both scripts?

pipelines/clusteranalysis/subworkflows/msa.nf
```diff
@@ -0,0 +1,159 @@
...
+}
+
+process muscle3_align {
+    tag "ca_muscle3_${type}_${id}"
+
+    publishDir "${params.final_output_dir}/data/${type}/msa", mode: "copy", pattern: "*.afa"
+
+    // MUSCLE can crash due to out of memory errors, or invalid data.  If we crash due to OOM,
+    // retry twice with larger amounts of RAM.  Otherwise, ignore the failure and proceed with
+    // the other clusters.
```
rbdavid, line 34: This seems like an implementation-specific nightmare. For example, when a job's allocation of memory is limited/set by the number of cores requested/allocated for the job.

pipelines/clusteranalysis/subworkflows/msa.nf
```diff
@@ -0,0 +1,159 @@
...
+    output:
+        path("*.zip")
+
+    script:
+    """
+    if [ -n "${clustal_pims}" ]; then
+        dir="clustal_pims_${type}"
+        mkdir \$dir
+        cp *.txt \$dir
+        zip -r "clustal_pims_${type}.zip" \$dir
```
rbdavid: This will save the files in the archive with the path, e.g. `${dir}/file_name.txt`. I just wanna check that we want to keep that directory structure w/in the archive.

pipelines/clusteranalysis/subworkflows/msa.nf
```diff
@@ -0,0 +1,159 @@
...
+
+        make_weblogos(msa_ch)
+
+        clustal_ch = run_clustal_omega(msa_ch)
+            .groupTuple()
+        zip_clustal_omega(clustal_ch)
+
+        if (params.pid_thresholds && !params.pid_thresholds.isEmpty()) {
+            //clustal_inputs = msa_ch.combine(Channel.fromList(params.pid_thresholds))
+            //clustal_inputs.view()
```
rbdavid: Should these be commented out right now?

pipelines/clusteranalysis/subworkflows/prepare.nf
```diff
@@ -0,0 +1,98 @@
...
+
+    output:
+        // We emit the same triplet structure so it can mix back easily later
+        tuple val(type), val(id), path("*_cdhit.fasta")
+
+    script:
+
+    """
+    # Using legacy parameters: -c 1 -s 1 -M 14900
+    cd-hit -c 1 -s 1 -i ${fasta} -o ${id}_cdhit.fasta -M 14900
```
rbdavid: `-M 14900` is a memory allocation max that should not be hardcoded. Use `params.cdhit_memory_limit`, which is defined in the `default_params.config` file.

pipelines/clusteranalysis/subworkflows/prepare.nf
```diff
@@ -0,0 +1,98 @@
...
+    output:
+        tuple val(type), val(id), path("*_subset.fasta")
+
+    script:
+
+    """
+wget https://github.com/shenwei356/seqkit/releases/download/v2.8.2/seqkit_linux_amd64.tar.gz
+    tar -xvf seqkit_linux_amd64.tar.gz
+    # --two-pass dramatically reduces memory usage for large files
+    # -s 100 sets a fixed seed for reproducibility
```
rbdavid: Should this be installed as a dependency? I don't think this is smart to include as a nextflow pipeline because, depending on implementation, some HPC resources won't allow internet connections while a job is running on a compute node.

pipelines/clusteranalysis/subworkflows/prepare.nf
```diff
@@ -0,0 +1,98 @@
...
+    """
+}
+
+workflow prepare_fasta {
+    take:
+        color_fasta_files
+
+    main:
+
+        // STEP 1: COLLECT ALL SEQUENCES INTO ONE CHANNEL
```
rbdavid, line 65: I don't really follow what this workflow is supposed to be doing. Can you point me to the old/legacy code that it is trying to recreate/replace?

rbdavid (commented):
Sorry for the disjointed reviews today. I have some general comments and questions that would help me during reviews if you answered. Thanks!

Dockerfile
rbdavid: What do you mean by "Yes"? Are you agreeing that making changes in distant parts of the code is not ideal when those changes could be more concentrated?

pipelines/shared/python/compute_length_histogram.py
rbdavid: I'm not sure of all the possible forms of the accession table file but, for a UniProt entry that doesn't have a UniRef90/50 label, wouldn't that line still have three tab separated columns in the file? Using zero-indexing, the 1st and 2nd columns would just be empty strings though, correct?

pipelines/clusteranalysis/subworkflows/msa.nf
nilsoberg: Fixed

pipelines/clusteranalysis/subworkflows/prepare.nf
nilsoberg: Yes, that was the plan all along.  I had to add things like this while testing due to an outdated docker build.

pipelines/clusteranalysis/logo/plot_hmm_weblogo.py
nilsoberg: I changed the code that generates the JSON files necessary for Skylign to render dynamic logos.  In the legacy code the Perl module `Bio::HMM::Logo` was used to generate these files.  It is not trivial to install in an automated fashion (e.g. Docker).  In a previous commit the code was using machine-generated code that didn't output the correct JSON.  Rather than write a brand-new script requiring an understanding of the nuances of both weblogos and the metadata necessary for Skylign, I machine-translated the Perl code into Python, and the result is the `pipelines/clusteranalysis/logo/BioHMMLogo.py` code.  Once we are happy with the result of the weblogo generation, the `BioHMMLogo.py` and `consensus_colors.py` files could/should get moved into `lib/pyEFI/pyEFI` as part of the EST Python package.  `convert_hmm_to_skylign_json.py` uses the `BioHMMLogo.py` code.

`plot_hmm_weblogo.py` has not been refactored (yet).  It is used to generate static images in a "wrapped" format.  `plot_hmm_weblogo.py` and `convert_hmm_to_skylign_json.py` do two different things and while it would be nice to share code between them, I don't think the effort involved in refactoring that is worth it.

pipelines/clusteranalysis/subworkflows/prepare.nf
nilsoberg: https://github.com/EnzymeFunctionInitiative/GNT/blob/master/lib/EFI/HMM/Job.pm

Dockerfile
nilsoberg: I updated the commands so that the path is set after every app install.

pipelines/shared/python/compute_length_histogram.py
nilsoberg: I think this is address two lines later where the ID is only added to the histogram list if the ID value is not empty.

pipelines/clusteranalysis/logo/plot_hmm_weblogo.py
nilsoberg: This file is no longer necessary because other scripts are used instead (`convert_hmm_to_skylign_json.py`)

pipelines/shared/python/compute_length_histogram.py
nilsoberg: I agree it would be better to use SeqIO and I'll keep it in mind for future refactors.  For now I want to stick with the existing code because I know it works.  This histogram code is used both in the `est` pipeline and `clusteranalysis` and I want to avoid making changes that could break `est`.

pipelines/shared/python/compute_length_histogram.py
nilsoberg: I believe this is an issue in the `est` pipeline as well.  Throwing an exception is one solution (or exiting non-zero).  The consequences of throwing an exception or non-zero though would cause the Nextflow pipeline to stop.  When generating length histograms for lots of clusters in the cluster analysis pipeline, we aren't worried about some missing (e.g. it shouldn't stop the job from completing).  We would either need to build logic into the process block to ignore errors or check the file size in bash before running the plotting script.

rbdavid: Could the code just skip writing the file so that, in the relevant process blocks, we can check for file existence to catch fatal instances where no sequences made it passed the filtering step?

nilsoberg:
`requirements.txt` needs to be updated to add PIL and pyhmmer.

pipelines/shared/perl/get_sequences.pl
rbdavid, line 124: Missing domain_id_map argument pod string.

pipelines/shared/python/compute_length_histogram.py
```diff
@@ -24,56 +33,94 @@ def parse_fasta(file_handle: TextIO) -> Iterator[Tuple[str, str]]:
...
     sequence_lengths = []
     try:
         with open(fasta_file, 'r') as fh:
             for header, sequence in parse_fasta(fh):
                 include_sequence = False
 
-                # Include the sequence if it starts with 'ZZ' (no UniProt match was found for the header)
+                # Include the sequence if it starts with 'ZZ' (e.g. the input to the EST pipeline
+                # was a FASTA file that didn't have UniProt IDs in the header).
                 if header.startswith('ZZ'):
```
rbdavid: Check blast input sequence id format. Might be a single "Z" character so catch that if that's the case.

pipelines/shared/python/compute_length_histogram.py
rbdavid: Make an issue for tracking this EST filtering check for zero sequences passing through the filtering.

pipelines/clusteranalysis/subworkflows/msa.nf
```diff
@@ -0,0 +1,315 @@
...
+}
+
+process collect_aa_ids {
+    tag "ca_count_msa_aa"
+
+    publishDir "${params.final_output_dir}/data/cons_res/${type}/consensus_residue_results_${aa}", mode: "copy", pattern: "*.txt"
+
+    input:
+        tuple path("msa_files/*"), path("logo_files/*"), path(cluster_size_file), path(mapping_file), val(aa), val(threshold), val(type)
+//        path("msa_files/*")
```
rbdavid: Outdated comments.

pipelines/clusteranalysis/subworkflows/msa.nf
```diff
@@ -0,0 +1,315 @@
...
+//        path id_cluster_mapping
+//        tuple val(aa), val(threshold)
+
+    output:
+        tuple val(aa), val(threshold), path("consensus_residue_${aa}_${threshold}_position.txt"), path("consensus_residue_${aa}_${threshold}_percentage.txt"), val(type)
+
+    script:
+    def base_name = "consensus_residue_${aa}_${threshold}"
+    """
+    perl $projectDir/conv_ratio/count_msa_aa.pl --msa-dir msa_files --logo-dir logo_files --aa ${aa} --count-file ${base_name}_position.txt --pct-file ${base_name}_percentage.txt --threshold ${threshold} --node-count-file ${cluster_size_file}
```
rbdavid: Follow consistent formatting of many-argument cli calls.

pipelines/clusteranalysis/logo/convert_hmm_to_skylign_json.py
```diff
@@ -0,0 +1,33 @@
+
+import argparse
+import json
+from BioHMMLogo import BioHMMLogo
```
rbdavid: Move into `lib/pyEFI` at some point.

rbdavid (approved):
Minor changes and clean up. We've discussed in person.

pipelines/shared/python/compute_length_histogram.py
nilsoberg: This will be fixed in #287

nilsoberg:
Closes #286

PR #278, opened 2026-02-17
Update shared color workflow to return SSN sequence as Nextflow value

nilsoberg:
The cluster analysis pipeline requires a value indicating the original sequence type that the SSN is built from (e.g. UniProt, UniRef90, UniRef50).  This PR has changes that return this value from the shared color workflow.

### DIFF
--- a/pipelines/shared/nextflow/color_workflow.nf
+++ b/pipelines/shared/nextflow/color_workflow.nf
@@ -64,11 +64,13 @@ process get_ssn_id_info {
         path "index_seqid_map.txt", emit: "index_seqid_map"     // Maps node network ID to UniProt ID and the number of IDs in the metanode
         path "seqid_source_map.txt", emit: "seqid_source_map"   // Maps metanode IDs to UniProt IDs
         path "ssn_sequences.fasta", emit: "ssn_sequences"       // Custom sequences that are embedded in the SSN
+        env SEQ_TYPE, emit: sequence_type                       // Type of sequences that the SSN is based on (uniprot, uniref90, uniref50)
 
     script:
     """
     perl $projectDir/../shared/perl/ssn_to_id_list.pl --ssn $ssn_file --edgelist edgelist.txt --index-seqid index_seqid_map.txt \
-        --seqid-source-map seqid_source_map.txt --ssn-sequences ssn_sequences.fasta
+        --seqid-source-map seqid_source_map.txt --ssn-sequences ssn_sequences.fasta --sequence-type-file sequence_type.txt
+    SEQ_TYPE=\$(cat sequence_type.txt)
     """
 }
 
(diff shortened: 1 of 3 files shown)


pipelines/shared/nextflow/color_workflow.nf
```diff
@@ -64,11 +64,13 @@ process get_ssn_id_info {
         path "index_seqid_map.txt", emit: "index_seqid_map"     // Maps node network ID to UniProt ID and the number of IDs in the metanode
         path "seqid_source_map.txt", emit: "seqid_source_map"   // Maps metanode IDs to UniProt IDs
         path "ssn_sequences.fasta", emit: "ssn_sequences"       // Custom sequences that are embedded in the SSN
+        env SEQ_TYPE, emit: sequence_type                       // Type of sequences that the SSN is based on (uniprot, uniref90, uniref50)
```
rbdavid, line 67: Why does this need to be emitted as an environment variable?

pipelines/shared/nextflow/color_workflow.nf
```diff
@@ -64,11 +64,13 @@ process get_ssn_id_info {
...
 
     script:
     """
     perl $projectDir/../shared/perl/ssn_to_id_list.pl --ssn $ssn_file --edgelist edgelist.txt --index-seqid index_seqid_map.txt \
-        --seqid-source-map seqid_source_map.txt --ssn-sequences ssn_sequences.fasta
+        --seqid-source-map seqid_source_map.txt --ssn-sequences ssn_sequences.fasta --sequence-type-file sequence_type.txt
+    SEQ_TYPE=\$(cat sequence_type.txt)
```
rbdavid, line 73: This `sequence_type.txt` is a file with a single line, correct?

pipelines/shared/nextflow/color_workflow.nf
```diff
@@ -270,6 +272,9 @@ workflow color_and_retrieve {
             .set { grouped_fasta_ch }
         zipped_fasta_dirs = zip_fasta_directories(grouped_fasta_ch)
 
+        // Convert to value channel
+        sequence_type_val = ssn_data.sequence_type.map { it.trim() }
```
rbdavid, line 276: I'm not sure why this needs to be `.map`'d since its an environment variable that is a bash string, not an array or anything that the nextflow `map` operator would handle (as I understand it). It probably works, but it just seems unnecessary.

rbdavid (approved):

pipelines/shared/nextflow/color_workflow.nf
nilsoberg: This is one of two ways to output a value in Nextflow (which is designed to work with files).  It gets converted into a value channel by Nextflow.  This value is required in later stages in cluster analysis (in a future PR) that will restrict computations to the input sequence type.  (For example, if the input SSN is a UniRef50 network, then the MSA and weblogo computations will only be done for the UniRef50 sequences.)

pipelines/shared/nextflow/color_workflow.nf
nilsoberg: Yes

pipelines/shared/nextflow/color_workflow.nf
nilsoberg: To remove newline characters

PR #281, opened 2026-02-17
Compute MSA and related processes using input SSN sequence type

nilsoberg:
- MSA, weblogo, HMM, and Clustal-Omega processes use input SSN sequence type (e.g. if SSN is UniRef50, UniRef50 sequences are used)
- In prior version all sequence types were processed
- Also, move computation of number of FASTA sequences into separate process; this previously occurred on the head node since the countFasta() function was in a workflow, so this was moved to a new process that counts the number of sequences

### DIFF
--- a/pipelines/clusteranalysis/subworkflows/msa.nf
+++ b/pipelines/clusteranalysis/subworkflows/msa.nf
@@ -5,10 +5,10 @@ process muscle5_align {
     publishDir "${params.final_output_dir}/data/${type}/msa", mode: "copy", pattern: "*.afa"
 
     input:
-        tuple val(type), val(id), path(fasta)
+        tuple val(type), val(id), path(fasta), val(seq_type), val(num_seq)
 
     output:
-        tuple val(type), val(id), path("*.afa"), emit: msa
+        tuple val(type), val(id), path("*.afa"), val(seq_type), val(num_seq), emit: msa
 
     script:
     """
@@ -38,10 +38,10 @@ process muscle3_align {
     }
 
     input:
-        tuple val(type), val(id), path(fasta)
+        tuple val(type), val(id), path(fasta), val(seq_type), val(num_seq)
 
     output:
-        tuple val(type), val(id), path("*.afa"), emit: msa
+        tuple val(type), val(id), path("*.afa"), val(seq_type), val(num_seq), emit: msa
 
     script:
     """
@@ -56,7 +56,7 @@ process build_hmms {
     publishDir "${params.final_output_dir}/data/${type}/hmms", mode: "copy", pattern: "*.json"
 
     input:
-        tuple val(type), val(id), path(msa)
+        tuple val(type), val(id), path(msa), val(seq_type), val(num_seq)
 
     output:
         path("*.hmm")
@@ -75,7 +75,7 @@ process make_weblogos {
     publishDir "${params.final_output_dir}/data/${type}/weblogos", mode: "copy", pattern: "*.png"
 
     input:
-        tuple val(type), val(id), path(msa)
+        tuple val(type), val(id), path(msa), val(seq_type), val(num_seq)
 
     output:
         path("*.png")
@@ -96,7 +96,7 @@ process run_clustal_omega {
     errorStrategy 'ignore'
 
     input:
-        tuple val(type), val(id), path(msa)
+        tuple val(type), val(id), path(msa), val(seq_type), val(num_seq)
 
     output:
         tuple val(type), path("*_pim.txt")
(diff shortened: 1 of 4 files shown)


pipelines/clusteranalysis/subworkflows/msa.nf
```diff
@@ -5,10 +5,10 @@ process muscle5_align {
     publishDir "${params.final_output_dir}/data/${type}/msa", mode: "copy", pattern: "*.afa"
 
     input:
-        tuple val(type), val(id), path(fasta)
+        tuple val(type), val(id), path(fasta), val(seq_type), val(num_seq)
 
     output:
-        tuple val(type), val(id), path("*.afa"), emit: msa
+        tuple val(type), val(id), path("*.afa"), val(seq_type), val(num_seq), emit: msa
```
rbdavid, line 11: These new outputs are also inputs and don't get new values during the process block. They aren't used in the process block either. Is the only reason we are passing them is to create the output tuple with those values included? It seems unnecessary.

pipelines/clusteranalysis/subworkflows/msa.nf
```diff
@@ -38,10 +38,10 @@ process muscle3_align {
     }
 
     input:
-        tuple val(type), val(id), path(fasta)
+        tuple val(type), val(id), path(fasta), val(seq_type), val(num_seq)
 
     output:
-        tuple val(type), val(id), path("*.afa"), emit: msa
+        tuple val(type), val(id), path("*.afa"), val(seq_type), val(num_seq), emit: msa
```
rbdavid, line 44: See above comment.

pipelines/clusteranalysis/subworkflows/msa.nf
```diff
@@ -56,7 +56,7 @@ process build_hmms {
     publishDir "${params.final_output_dir}/data/${type}/hmms", mode: "copy", pattern: "*.json"
 
     input:
-        tuple val(type), val(id), path(msa)
+        tuple val(type), val(id), path(msa), val(seq_type), val(num_seq)
```
rbdavid, line 59: These are just added as inputs so that you can feed the output from the above processes into this process without explicitly calling variables in the tuple?

pipelines/clusteranalysis/subworkflows/msa.nf
```diff
@@ -75,7 +75,7 @@ process make_weblogos {
     publishDir "${params.final_output_dir}/data/${type}/weblogos", mode: "copy", pattern: "*.png"
 
     input:
-        tuple val(type), val(id), path(msa)
+        tuple val(type), val(id), path(msa), val(seq_type), val(num_seq)
```
rbdavid, line 78: See above.

pipelines/clusteranalysis/subworkflows/prepare.nf
```diff
@@ -3,31 +3,45 @@ process subset_fasta {
...
 
     script:
 
     """
-wget https://github.com/shenwei356/seqkit/releases/download/v2.8.2/seqkit_linux_amd64.tar.gz
-    tar -xvf seqkit_linux_amd64.tar.gz
     # --two-pass dramatically reduces memory usage for large files
     # -s 100 sets a fixed seed for reproducibility
-    ./seqkit sample --two-pass -n ${params.max_msa_sequences} -s 100 ${fasta} > ${id}_subset.fasta
+    seqkit sample --two-pass -n ${params.max_msa_sequences} -s 100 ${fasta} > ${id}_subset.fasta
```
rbdavid, line 16: Should the random seed be defined in the `default_params.config` file and overwritten by a user-set parameter (or random number generator) when a real random seed is desired to be used?

Alternatively, I think the point of the random seed is not to keep it constant for all analyses being run via seqkit. But, instead, to report what the randomly generated seed was so that scientists can reproduce the results after the fact. Using the same seed for all calculations is not the intended usage of a random seed.

pipelines/clusteranalysis/subworkflows/prepare.nf
rbdavid, line 1: I still don't follow what this prepare workflow is doing. Could you point me to the legacy code that its refactoring?

rbdavid (approved):

pipelines/clusteranalysis/subworkflows/msa.nf
nilsoberg: Yes

nilsoberg: It's needed for consistency between processes

pipelines/clusteranalysis/subworkflows/prepare.nf
nilsoberg: https://github.com/EnzymeFunctionInitiative/GNT/blob/master/lib/EFI/HMM/Job.pm

PR #284, opened 2026-02-23
283 - Fix duckdb temp dir naming

### LINKED ISSUE #283: task.hash evaluates to null
Lines in the `all_by_all.nf` workflow script that use the `--duckdb-temp-dir ${params.duckdb_temp_dir}-${task.hash}` are not functioning as expected. [Nextflow documentation](https://www.nextflow.io/docs/latest/reference/process.html#task-properties) says that `task.hash` is only usable in `exec` blocks, otherwise it evaluates to `null`. This is seen when looking at the process block's `.command.sh` scripts associated with the duckdb calls: 

```
work/fc/9e51d1acc18243f48c618080446fd3/.command.sh:python EST/pipelines/est/axa_blast/render_prereduce_sql_template.py --blast-output fracfile-43.fa.tab.parquet --sql-template EST/pipelines/est/templates/prereduce-template.sql --output-file fracfile-43.fa.tab.sorted.parquet --duckdb-memory-limit 8GB --duckdb-temp-dir /scratch/duckdb-null --sql-output-file prereduce.sql
```

where `--duckdb-temp-dir ${params.duckdb_temp_dir}-${task.hash}` evaluates to `--duckdb-temp-dir /scratch/duckdb-null`. This does not throw any errors and, as far as I can find, any warnings about the undefined `task.hash` namespace.

These lines need to be updated to use a randomized or sufficiently rare identifier so that two running processes won't use the same temp duckdb storage space.

rbdavid:
- [x] Changed `task.hash` to `task.index` to avoid the `exec`-block only hash string, which evaluates to null within a process block. `task.index` uses the process block's index in the workflow (AFAIK) and so should be unique for the given job. But, if multiple jobs are running on the same compute node, this may cause an issue.
- [x] The `generatessn.nf` code has a call to duckdb and an associated template file that accepts memory limit and temp directory definitions. I've implemented those input arguments to the rendering python script so that they are controlled by the implementation specific parameters rather than left up to default values set by duckdb.

### DIFF
--- a/lib/pyEFI/pyEFI/sql_template_render.py
+++ b/lib/pyEFI/pyEFI/sql_template_render.py
@@ -1,10 +1,28 @@
 import argparse
 from io import TextIOWrapper
 import string
+import tempfile
 
 from typing import Any
 
-def create_sql_template_render_parser(sql_template_file_default: str, desc: str, sql_output_file="statements.sql", duckdb_mem_limit="4GB", duckdb_temp_dir="./duckdb") -> argparse.ArgumentParser:
+def get_temp_dir_name() -> str:
+    """
+    Use tempfile.TemporaryDirectory() to create a temp directory space, gather
+    the path for that space, and automatically delete the empty directory.
+    Return the path to duckdb during sql template rendering.
+    """
+    temp_dir = tempfile.TemporaryDirectory(ignore_cleanup_errors = True)
+    temp_path = temp_dir.name
+    temp_dir.cleanup()
+    return temp_path
+
+def create_sql_template_render_parser(
+        sql_template_file_default: str,
+        desc: str,
+        duckdb_temp_dir: str,
+        sql_output_file: str = "statements.sql",
+        duckdb_mem_limit: str = "4GB"
+    ) -> argparse.ArgumentParser:
     """
     Returns an `<argparse.ArgumentParser>_` that parses the following
     options:
@@ -44,7 +62,9 @@ def create_sql_template_render_parser(sql_template_file_default: str, desc: str,
     parser.add_argument(
         "--duckdb-temp-dir",
         type=str,
-        default=duckdb_temp_dir,
+        # if left undefined, will create a randomly generated directory name in
+        # the system's defined temp space
+        default = get_temp_dir_name(),
         help="Location DuckDB should use for temporary files",
     )
     return parser
--- a/lib/pyEFI/pyEFI/sql_template_render.py
+++ b/lib/pyEFI/pyEFI/sql_template_render.py
@@ -2,19 +2,17 @@ import argparse
 from io import TextIOWrapper
 import string
 import tempfile
+import uuid
 
 from typing import Any
 
-def get_temp_dir_name() -> str:
+def get_temp_dir_name(tmp_path: str = tempfile.gettempdir()) -> str:
     """
     Use tempfile.TemporaryDirectory() to create a temp directory space, gather
     the path for that space, and automatically delete the empty directory.
     Return the path to duckdb during sql template rendering.
     """
-    temp_dir = tempfile.TemporaryDirectory(ignore_cleanup_errors = True)
-    temp_path = temp_dir.name
-    temp_dir.cleanup()
-    return temp_path
+    return os.path.join(tmp_path, "duckdb-" + str(uuid.uuid4())[:8])
 
 def create_sql_template_render_parser(
         sql_template_file_default: str,
--- a/lib/pyEFI/pyEFI/sql_template_render.py
+++ b/lib/pyEFI/pyEFI/sql_template_render.py
@@ -1,24 +1,26 @@
 import argparse
 from io import TextIOWrapper
 import string
-import tempfile
-import uuid
+
+#import tempfile
+#import uuid
 
 from typing import Any
 
-def get_temp_dir_name(tmp_path: str = tempfile.gettempdir()) -> str:
-    """
-    Use tempfile.TemporaryDirectory() to create a temp directory space, gather
-    the path for that space, and automatically delete the empty directory.
-    Return the path to duckdb during sql template rendering.
-    """
-    return os.path.join(tmp_path, "duckdb-" + str(uuid.uuid4())[:8])
+#def get_temp_dir_name(tmp_path: str = ) -> str:
+#    """
+#    Use tempfile.TemporaryDirectory() to create a temp directory space, gather
+#    the path for that space, and automatically delete the empty directory.
+#    Return the path to duckdb during sql template rendering.
+#    """
+#    return os.path.join(tmp_path, "duckdb-" + str(uuid.uuid4())[:8])
 
 def create_sql_template_render_parser(
         sql_template_file_default: str,
         desc: str,
         sql_output_file: str = "statements.sql",
-        duckdb_mem_limit: str = "4GB"
+        duckdb_mem_limit: str = "4GB",
+        duckdb_temp_dir: str = "./duckdb"
     ) -> argparse.ArgumentParser:
     """
     Returns an `<argparse.ArgumentParser>_` that parses the following
@@ -59,9 +61,7 @@ def create_sql_template_render_parser(
     parser.add_argument(
         "--duckdb-temp-dir",
         type=str,
-        # if left undefined, will create a randomly generated directory name in
-        # the system's defined temp space
-        default = get_temp_dir_name(),
+        default = duckdb_temp_dir,
         help="Location DuckDB should use for temporary files",
     )
     return parser
--- a/lib/pyEFI/pyEFI/sql_template_render.py
+++ b/lib/pyEFI/pyEFI/sql_template_render.py
@@ -19,7 +19,6 @@ def get_temp_dir_name() -> str:
 def create_sql_template_render_parser(
         sql_template_file_default: str,
         desc: str,
-        duckdb_temp_dir: str,
         sql_output_file: str = "statements.sql",
         duckdb_mem_limit: str = "4GB"
     ) -> argparse.ArgumentParser:
--- a/lib/pyEFI/pyEFI/sql_template_render.py
+++ b/lib/pyEFI/pyEFI/sql_template_render.py
@@ -2,26 +2,9 @@ import argparse
 from io import TextIOWrapper
 import string
 
-#import tempfile
-#import uuid
-
 from typing import Any
 
-#def get_temp_dir_name(tmp_path: str = ) -> str:
-#    """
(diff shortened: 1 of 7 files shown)


rbdavid:
closes #283

nilsoberg (changes requested):
By using the process ID (`$$`), you can construct a completely unique directory. 

```
    DUCKDB_TEMP="${params.duckdb_temp_dir}/duckdb-${task.index}-\$\$"
    python $projectDir/axa_blast/render_prereduce_sql_template.py --blast-output ${frac}.tab.parquet --sql-template $projectDir/templates/prereduce-template.sql --output-file ${frac}.tab.sorted.parquet --duckdb-memory-limit ${params.duckdb_memory_limit} --duckdb-temp-dir \$DUCKDB_TEMP --sql-output-file prereduce.sql
```

rbdavid:
I think, instead, we should shift the generation of a randomly named temp_directory to within the python scripts that render the sql query files. 

The `$$`, `${PID}`, and any other form of bash command/function to get a semi-randomized string fails because the nextflow `script` strings for these process blocks all use `"""` instead of `'''`, which controls how `$`'d variables are handled (see [documentation](https://training.nextflow.io/2.0/basic_training/processes/#script-parameters)). If double quotes are used, nextflow interprets all `$variables` calls within the block as nextflow variables and raises fatal errors about, e.g. `$$`, not being a defined variable in the nextflow environment. If, instead, we used `'''` around the script block, all of our `$frac` or `params.input_argument` calls fail because those are not known by the bash environment.

Instead, the python scripts can use something like:
``` python
import tempfile
temp_dir = tempfile.mkdtemp()
print(temp_dir)

>> /tmp/tmpnskma309
(rerunning these lines)
>> /tmp/tmpofm5m3xl
```
Having a randomly named temp_directory seems to be very important because duckdb follows a fairly rigid temporary file naming scheme, so if two subprocesses are writing temp files to the same temp directory there will be clashes. I think the python script above is the best means to create the required temp storage space.

lib/pyEFI/pyEFI/sql_template_render.py
```diff
@@ -1,10 +1,28 @@
...
-def create_sql_template_render_parser(sql_template_file_default: str, desc: str, sql_output_file="statements.sql", duckdb_mem_limit="4GB", duckdb_temp_dir="./duckdb") -> argparse.ArgumentParser:
+def get_temp_dir_name() -> str:
+    """
+    Use tempfile.TemporaryDirectory() to create a temp directory space, gather
+    the path for that space, and automatically delete the empty directory.
+    Return the path to duckdb during sql template rendering.
+    """
+    temp_dir = tempfile.TemporaryDirectory(ignore_cleanup_errors = True)
+    temp_path = temp_dir.name
+    temp_dir.cleanup()
```
rbdavid: Instantiating the TemporaryDirectory object creates the associated directory in the storage. But, we don't want that directory to actually exist when we pass it to duckdb because that can send mixed signals to duckdb about whether the temp directory space should be cleaned up or not (https://github.com/duckdb/duckdb/issues/5878; duckdb documentation on how the temp_directory is handled upon context closing is a bit ambiguous). So, to be safe, just let duckdb create the temp_dir space and clean it up.

rbdavid:
With this latest change, the rendered sql scripts now contain randomized temp directories with clean up handled by duckdb. 

```
$ grep "temp_directory" work/*/*/*sql
work/01/198dbdf849ac705915a83d16e3b99c/prereduce.sql:SET temp_directory = '/tmp/tmpargtzpw5';
work/05/be69fcbc2ae05de8a0f9560435f163/prereduce.sql:SET temp_directory = '/tmp/tmp5b25m0iw';
work/1e/98f540c9a863507168f14340e59927/prereduce.sql:SET temp_directory = '/tmp/tmptm03wru4';
...
```

rbdavid:
Only one instance of calling `params.duckdb_temp_dir` remains in the current workflows. Specifically, in `restore_condensed` process block in the `all_by_all.nf` workflow code. This process block does not use the sql template rendering python script as other processes do. This process block is run in serial (so should not clash with other process blocks running from the same workflow). But, there is still the possibility that a different workflow job is running on the same compute resource, and the running workflow instances have overlapping times where `restore_condensed` process blocks are running, which likely would result in duckdb temp file clashes.

rbdavid:
I've implemented a `render_restore_sql_template.py` and associated sql template file to standardize all of the duckdb interfaces/calls in the current workflows.

nilsoberg (commented):
I like the idea that we don't have to track another nextflow parameter (by removing `duckdb_temp_dir`).  However, the system temp dir may not have much space.  Additionally, I suspect that most HPC use a dedicated scratch partition with high speed storage.  In one configuration that I'm directly involved with I believe `/tmp` is mounted on a slow disk and `/scratch` is dedicated, fast NVME storage.  If this is the case then we would need to at least provide the user with the ability to specify the temp dir (perhaps defaulting to the system temp if not provided).  @dslater82 could you confirm this architecture?

dslater82:
/scratch is fast disk and /tmp is on the node image which itself is in memory but with very limited space.  Just enough for the OS to work.  We always recommend making the scratch space be /scratch/$SLURM_JOB_ID so it would be unique for each job.  You do not want temp files intermixing between different jobs.  Sometimes programs only look at $TMPDIR variable which by default goes to /tmp.  We recommend doing "export TMPDIR=/scratch/$SLURM_JOB_ID".

nilsoberg:
> I think, instead, we should shift the generation of a randomly named temp_directory to within the python scripts that render the sql query files.
> 
> The `$$`, `${PID}`, and any other form of bash command/function to get a semi-randomized string fails because the nextflow `script` strings for these process blocks all use `"""` instead of `'''`, which controls how `$`'d variables are handled (see [documentation](https://training.nextflow.io/2.0/basic_training/processes/#script-parameters)). If double quotes are used, nextflow interprets all `$variables` calls within the block as nextflow variables and raises fatal errors about, e.g. `$$`, not being a defined variable in the nextflow environment. If, instead, we used `'''` around the script block, all of our `$frac` or `params.input_argument` calls fail because those are not known by the bash environment.

I use `\$VARIABLE` all the time in the `script` section without issue.  In the Nextflow documentation you referenced, they give this example:

```
    echo "The current directory is \$PWD"
```

Notice the backslashes in the example I gave:

```
    DUCKDB_TEMP="${params.duckdb_temp_dir}/duckdb-${task.index}-\$\$"
    python $projectDir/axa_blast/render_prereduce_sql_template.py --blast-output ${frac}.tab.parquet --sql-template $projectDir/templates/prereduce-template.sql --output-file ${frac}.tab.sorted.parquet --duckdb-memory-limit ${params.duckdb_memory_limit} --duckdb-temp-dir \$DUCKDB_TEMP --sql-output-file prereduce.sql
```

That would render to the `.command.sh` script as:

```
DUCKDB_TEMP="SOMEDIR/duckdb-#-$$"
python pipelines/est/axa_blast/render_prereduce_sql_template.py --blast-output 1.tab.parquet --sql-template pipelines/est/templates/prereduce-template.sql --output-file 1.tab.sorted.parquet --duckdb-memory-limit 10G --duckdb-temp-dir $DUCKDB_TEMP --sql-output-file prereduce.sql
```

rbdavid:
> I like the idea that we don't have to track another nextflow parameter (by removing `duckdb_temp_dir`). However, the system temp dir may not have much space. Additionally, I suspect that most HPC use a dedicated scratch partition with high speed storage. In one configuration that I'm directly involved with I believe `/tmp` is mounted on a slow disk and `/scratch` is dedicated, fast NVME storage. If this is the case then we would need to at least provide the user with the ability to specify the temp dir (perhaps defaulting to the system temp if not provided). @dslater82 could you confirm this architecture?

I'll add back the `params.duckdb_temp_dir` and re-implement its usage in the python scripts arguments.

rbdavid:
> I use `\$VARIABLE` all the time in the `script` section without issue. In the Nextflow documentation you referenced, they give this example:
> 
> ```
>     echo "The current directory is \$PWD"
> ```
> 
> Notice the backslashes in the example I gave:
> 
> ```
>     DUCKDB_TEMP="${params.duckdb_temp_dir}/duckdb-${task.index}-\$\$"
>     python $projectDir/axa_blast/render_prereduce_sql_template.py --blast-output ${frac}.tab.parquet --sql-template $projectDir/templates/prereduce-template.sql --output-file ${frac}.tab.sorted.parquet --duckdb-memory-limit ${params.duckdb_memory_limit} --duckdb-temp-dir \$DUCKDB_TEMP --sql-output-file prereduce.sql
> ```
> 
> That would render to the `.command.sh` script as:
> 
> ```
> DUCKDB_TEMP="SOMEDIR/duckdb-#-$$"
> python pipelines/est/axa_blast/render_prereduce_sql_template.py --blast-output 1.tab.parquet --sql-template pipelines/est/templates/prereduce-template.sql --output-file 1.tab.sorted.parquet --duckdb-memory-limit 10G --duckdb-temp-dir $DUCKDB_TEMP --sql-output-file prereduce.sql
> ```

I've tested the `\$\$` and other equivalent methods of creating a randomized dir name via bash commands. Specifically, the below lines are created from a test where I set the argument to `--duckdb-temp-dir /tmp/duckdb-${task.index}-\$\$`. 
```
$ grep "temp_directory" work/*/*/*sql
work/00/52075756f88cab710fdcac6dd4c2a3/prereduce.sql:SET temp_directory = '/tmp/duckdb-25-26';
work/02/58b389ae6efe2eb78e4b704e671c48/prereduce.sql:SET temp_directory = '/tmp/duckdb-21-26';
work/1c/830bf319b0d67a6b877255f84a639c/prereduce.sql:SET temp_directory = '/tmp/duckdb-40-26';
work/1e/787e2e45e786bb9d6edb34eb4ec39b/prereduce.sql:SET temp_directory = '/tmp/duckdb-5-26';
...
```

The `\$\$` portion of the temp_dir value is static across all the subprocesses because its determined for the process block not for each instance of the process block. The all_by_all_blast process spins up 128 instances but they all get the same `$$` value (26). The combo with `${task.index}-\$\$` does not work perfectly either, because its possible that multiple different nextflow workflows are running on the same compute node. If two nextflow workflows are running on the same compute resource, both will assign the same temp dir space to one of the all_by_all_blast subprocesses, e.g. `/tmp/duckdb-25-26`. The temp directory name will be the exact same across those multiple jobs and we'll get temp directory/file clashes when multiple duckdb calls go to write in the temp_dir space. So, we truly needed to implement the semi/fully random directory naming scheme that I implemented in the PR.

nilsoberg (approved):
Looks great

PR #292, opened 2026-03-30
291 - Create post filter script for stats

### LINKED ISSUE #291: Add post-filter processing script to Taxonomy Tool
- [x] Create a python script to read import_stats.json, process those values to calculate more complete set of results stats. Needs to calculate a `num_unique_ids` key and write out the intermediary json file. 
- [x] Create the process block to run the above python script and return the file path. 
- [x] Feed that file path into the `merge_stats` process block.

rbdavid:
Create a separate process block to calculate the `num_unique_ids` value to be placed in `stats.json` file. Also fix placement of the validation for seq length filter values. 

Currently investigating a off-by-one issue that I'm observing in some of my taxonomy tool tests. Once that's figured out, this should be ready to review.  

Closes #291


rbdavid:
One thing that I'm noticing, is that for any type of taxonomy job, the `num_filter_length` key is always present in the `stats.json` file, even when that filter is not applied or not relevant for the taxonomy job type. I think it will be present even for EST jobs that run the `filter_ids` process block. I think this means that I messed up some of the logic about what values of the `min_seq_length` and `max_seq_length` filters should indicate that the length filter is actually applied.

lib/EFI/Import/Config/Filter.pm
```diff
@@ -73,8 +73,8 @@ sub validateOptions {
     $opts->{fraction} = ($filter->{fraction} and $filter->{fraction} =~ m/^\d+$/) ? $filter->{fraction} : 1;
     $opts->{remove_fragments} = exists $filter->{fragments} ? 1 : 0;
     $opts->{family_filter} = $filter->{family};
-    $opts->{min_seq_length} = $filter->{min_seq_length};
-    $opts->{max_seq_length} = $filter->{max_seq_length};
+    $opts->{min_seq_length} = ($filter->{min_seq_length} and $filter->{min_seq_length} =~ m/^\d+$/) ? $filter->{min_seq_length} : 1;
+    $opts->{max_seq_length} = ($filter->{max_seq_length} and $filter->{max_seq_length} =~ m/^\d+$/) ? $filter->{max_seq_length} : 65000;
```
nilsoberg: I would move these two values to a constant at the top of the file (e.g. `use constant DEFAULT_MIN_SEQ_LENGTH => 1;`).

pipelines/taxonomy/statistics/update_import_stats.py
nilsoberg, line 1: Maybe name this `update_import_stats.py`? Just a thought. The other thing that came to mind is the JSON files can be large (many megabytes).  Will this result in excessive RAM usage?

nilsoberg (approved):
This looks good.  Maybe move the default length values to a constant before merging.

nilsoberg:
> One thing that I'm noticing, is that for any type of taxonomy job, the `num_filter_length` key is always present in the `stats.json` file, even when that filter is not applied or not relevant for the taxonomy job type. I think it will be present even for EST jobs that run the `filter_ids` process block. I think this means that I messed up some of the logic about what values of the `min_seq_length` and `max_seq_length` filters should indicate that the length filter is actually applied.

I think that this is because the filter is always run.  It should only be run when the min and/or max length values are specified (even if they're default).  That's how the other filters work.

nilsoberg (approved):
Looks good

PR #298, opened 2026-04-13
Refactor color SSN workflow for modularity

### LINKED ISSUE #297: Refactor color workflow for modularity
Refactor the color workflow code for modularity.  Expose a workflow that can be used to generate SSNs as part of the convergence ratio, generate SSN, color SSN, cluster analysis, and neighborhood connectivity, rather than requiring colored SSNs as input.

nilsoberg:
- Separate ID files, FASTA retrieval from SSN cluster computation
- Main color workflow calls SSN cluster computation workflow
- SSN cluster computation workflow can be used by other workflows to color a SSN before running other workflows

### DIFF
--- a/pipelines/shared/nextflow/color_workflow.nf
+++ b/pipelines/shared/nextflow/color_workflow.nf
@@ -12,23 +12,23 @@ process get_id_list {
         val sequence_type
 
     output:
-        path "cluster_sizes.txt", emit: "cluster_sizes"
+        path "cluster_sizes.txt", emit: cluster_sizes
 
         // Outputs for FASTA retrieval
-        tuple val("uniprot"), path("uniprot/*.txt"), emit: "uniprot_tuples"
-        tuple val("uniref90"), path("uniref90/*.txt", arity: "0..*"), emit: "uniref90_tuples"
-        tuple val("uniref50"), path("uniref50/*.txt", arity: "0..*"), emit: "uniref50_tuples"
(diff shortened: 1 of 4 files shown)


pipelines/shared/nextflow/color_workflow.nf
```diff
@@ -255,26 +255,45 @@ process zip_fasta_directories {
...
+workflow COLOR_AND_RETRIEVE {
+    take:
+        ssn_file
+
+    main:
+        //
+        // STEP 1: PARSE THE SSN
+        //
+
+        computed = COMPUTE_COLOR_CLUSTER_WORKFLOW(ssn_file)
```
rbdavid, line 296: So is this the only instance of this `COMPUTE_COLOR_CLUSTER_WORKFLOW` workflow being called? Will there be other instances where just the work in that workflow is to be done (without any further steps in the `COLOR_AND_RETRIEVE` workflow)?

rbdavid (approved):
Other than the conceptual question I have, everything looks good. Its all rather minor changes if I understand correctly.

pipelines/shared/nextflow/color_workflow.nf
nilsoberg: One reason for this redesign was to allow `generatessn`, `convergenceratio`, `neighborhoodconnectivity`, and CGFP to dynamically create a colored/numbered SSN without requiring the user to run the `colorssn` pipeline beforehand.  `COMPUTE_COLOR_CLUSTER_WORKFLOW` can be called by all of those in the future.

PR #299, opened 2026-04-14
Add fragment and taxonomy filtering to `generatessn` pipeline

nilsoberg:
This PR includes changes to the `generatessn` pipeline to support fragment and taxonomy filters.

* Terminology was changed (previously "filter" referred to alignment score cutoff value; this is now "threshold")
* For clarity, provide arguments to `filter_ids.pl` in `pipelines/shared/nextflow/sequence.nf` -- previously they were implicitly defined
* Update default maximum length from 50,000 to 65,000 to match legacy site values
* Add additional filter type for `filter_ids.pl`: option to manually specify IDs to remove ("explicit IDs"); this is needed because of the length filters that are applied as part of the `compute_fasta_lengths ` (formerly `filter_fasta`) process.  The length filtering is done based on the actual sequence length not the full sequence length defined in the database.  Relevant in the case of an Option C FASTA job or domain-based sequences.
* `filter_fasta` was replaced with `compute_fasta_lengths` to compute the file for exclusion ("explicit IDs") and uses `seqkit` to compute the sequence lengths, removing the need for the Perl script.
* Add support for `COMPUTE_COLOR_CLUSTER_WORKFLOW` in `pipelines/shared/nextflow/color_workflow.nf` -- for coloring a SSN after it is created.


docs/source/pipelines/generatessn/index.rst
```diff
@@ -15,7 +15,7 @@ pipeline in an "automatic" mode, or the specific necessary files in "manual"
 mode.  The parameter file necessary to run the pipeline is created with the
 ``bin/create_generatessn_nextflow_params.py`` script: ::
 
-    python bin/create_generatessn_nextflow_params.py auto --filter-min-val 23 --ssn-name ssn.xgmml --ssn-title "SSN Title" --est-output-dir est_results/ --efi-config efi.config --efi-db efi_202406 --nextflow-config file.config
+    python bin/create_generatessn_nextflow_params.py auto --threshold-min-val 23 --ssn-name ssn.xgmml --ssn-title "SSN Title" --est-output-dir est_results/ --efi-config efi.config --efi-db efi_202406 --nextflow-config file.config
```
rbdavid, line 18: Might be worthwhile to also call `--threshold-metric` to remove ambiguity in this first example.

tests/modules/20c_est_generatessn_color_ssn.sh
rbdavid, line 1: Instead of re-running the EST process, could you just copy the result files and run the generatessn.nf process with the `--color-ssn` parameter?

tests/modules/20b_est_generate_ssn_taxonomy_filter.sh
rbdavid, line 1: Can you just copy the results of a past job instead of re-running the EST?

tests/modules/20a_est_generatessn_fragment_filter.sh
rbdavid, line 1: Can you just re-use the results from a past job?

rbdavid (approved):
I haven't run the new tests yet but the code seems fine.

rbdavid:
Also, to note, this is merging the 236 branch into nextflow-test. Is that intended? The branch (`236-generatessn-filtering-wip-taxonomy`) has "wip-taxonomy" in it, so should it be merged into `236-generatessn-filtering`. Maybe I'm getting confused about the three `generatessn` PRs that are currently open or in draft form.

pipelines/generatessn/threshold/render_threshold_blast_sql_template.py
```diff
@@ -10,20 +10,20 @@ def add_custom_arguments(parser: argparse.ArgumentParser) -> None:
...
         required=True,
-        help="The final output file the filtered BLAST output should be written to. Will be tab-separated"
+        help="The final output file the thresholded BLAST output should be written to. Will be tab-separated"
     )
     parser.add_argument(
-        "--filter-parameter",
+        "--threshold-metric",
         type=str,
         required=True,
         choices=["pident", "bitscore", "alignment_score"],
```
rbdavid, line 19: Just to note, this is the line where one of the three different metrics can be selected and applied in the SQL template file.

PR #306, opened 2026-04-17
Save the ssn_sequence_type stats value in lowercase for color SSN jobs

### LINKED ISSUE #285: Lowercase ColorSSN's colorSsnSequenceSource value
When writing the stats.json file, certain values are ideally written in lowercase format so that they cleanly map to efi-web enums. One such example is the key colorSsnSequenceSource's value.

nilsoberg:


rbdavid (approved):
The job_manager code handles the lower-casing of this string, so if it needs to be in "UniRef" or "UniProt" form for the rest of the backend codes, that's fine.

PR #308, opened 2026-04-24
287 support domain sequences in length histogram computation

### LINKED ISSUE #287: Support domain sequences in length histogram computation
The length histogram computation script currently doesn't support sequence IDs with domain boundaries, resulting in an empty histogram file.  When parsing the FASTA file, the domain information should be removed from the header ID to match the input accession list.

nilsoberg:
This PR updates the length histogram computation logic to support domain sequences.

* Modified the sequence length extraction scripts to correctly handle and count domain-specific sequence lengths
* Stop pipeline early in workflow if no sequences were found
* Improved naming of variables

Closes #287

### DIFF
--- a/pipelines/est/subworkflows/import.nf
+++ b/pipelines/est/subworkflows/import.nf
@@ -26,9 +26,17 @@ process cat_fasta_files {
         """
         $cat_cmd
         perl $projectDir/import/append_blast_query.pl --blast-query-file ${params.input_file} --output-sequence-file all_sequences.fasta
+
+        # Stop Nextflow here if the file is empty (i.e. no sequences were found)
+        [ -s all_sequences.fasta ] || { echo "ERROR: No sequences found after retrieval and merge."; exit 1; }
         """
     } else {
-        cat_cmd
+        """
+        $cat_cmd
+
+        # Stop Nextflow here if the file is empty (i.e. no sequences were found)
+        [ -s all_sequences.fasta ] || { echo "ERROR: No sequences found after retrieval and merge."; exit 1; }
+        """
     }
 }
 
--- a/pipelines/est/subworkflows/import.nf
+++ b/pipelines/est/subworkflows/import.nf
@@ -82,7 +82,7 @@ workflow IMPORT_AND_FILTER {
         // from the BLAST sequence database.  If the import mode is FASTA, then these IDs are only
         // ones that come from adding a family to the job
         accession_shards = split_sequence_ids(sequence_id_files.retrieval_ids, params.num_accession_shards)
-        fasta_files = get_sequences(accession_shards.flatten(), params.fasta_db)
+        raw_fasta_files = get_sequences(accession_shards.flatten(), params.fasta_db)
 
         // If importing FASTA file, reformat the FASTA file and create the file that will be added to
         // the dataset for all-by-all BLAST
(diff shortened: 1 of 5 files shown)


pipelines/est/subworkflows/import.nf
```diff
@@ -26,9 +26,17 @@ process cat_fasta_files {
         """
         $cat_cmd
         perl $projectDir/import/append_blast_query.pl --blast-query-file ${params.input_file} --output-sequence-file all_sequences.fasta
+
+        # Stop Nextflow here if the file is empty (i.e. no sequences were found)
+        [ -s all_sequences.fasta ] || { echo "ERROR: No sequences found after retrieval and merge."; exit 1; }
```
rbdavid, line 31: Could we use this PR as a first dive into error handling of specific process blocks? Instead of `exit 1`, could we specify a unique ID for the `cat_fasta_files` process block and use that ID when exiting due to error? That would ease the detection of what process block failed and reporting informative messaging back to users/devs.

rbdavid: This is just an idea. I still have on my to-do list a nextflow.log parsing script to gather this information. A unique exit code for each process block (or for each unique reason why nextflow error'd) would be helpful still, I think.

pipelines/est/subworkflows/import.nf
```diff
@@ -74,18 +82,24 @@ workflow IMPORT_AND_FILTER {
...
+            fasta_files = raw_fasta_files.concat(import_fasta_file)
+        } else {
+            fasta_files = raw_fasta_files
         }
 
-        fasta_file = cat_fasta_files(fasta_files.collect())
+        all_fasta_files = fasta_files.collect()
+        fasta_file = cat_fasta_files(
+            all_fasta_files
+                .ifEmpty { error "No FASTA sequences were retrieved. Terminating pipeline." }
```
rbdavid, line 101: Similar as above. Could we add an error code along with the messaging? Like, `error {unique_id}, "No FASTA sequences were retrieved. Terminating pipeline."`

pipelines/shared/nextflow/sequence.nf
```diff
@@ -128,19 +128,3 @@ process get_sequences {
     """
 }
 
-process get_length_histogram {
```
rbdavid, line 131: I believe my taxonomy tool work was expecting this `get_length_histogram` process block to be in `shared/nextflow/` space. Is there a reason this is being moved back into the `reporting.nf` code?

rbdavid (approved):
Good call on using Biopython fasta parsing tech. I think the home-made fasta parsing code was technical debt and handing it off to the module is for the best. Likely not for this PR but we should develop some pytest handles for the pyEFI codes to ensure, for example, biopython fasta parsing is doing what we need it to be doing. 

Other note is the placement of the `get_length_histograms` process block. Taxonomy tool needs to plot histograms so could that block remain in the shared nextflow codes?

pipelines/shared/nextflow/sequence.nf
nilsoberg: Correct, this was erroneously removed from the shared nextflow code.  I have added it back in.

pipelines/est/subworkflows/import.nf
nilsoberg: I like this idea.  If it's OK, I'll create an issue for it and mark it as a low priority.  We can co-develop the feature when you start on the log parsing script.

PR #310, opened 2026-04-24
309 make user taxonomy filter json format consistent

### LINKED ISSUE #309: Make user taxonomy filter JSON format consistent
User defined taxonomy filters are stored as hashes in the efi-web database.  Update the user defined taxonomy filtering code to support a hash rather than requiring the user defined filter to be nested in an array.

nilsoberg:
This PR standardizes the JSON format used for the user taxonomy filter across the EST toolset. Previously, there were inconsistencies in how taxonomy data was structured between the frontend input and the backend processing scripts, which could lead to issues during job submission or pipeline execution.

* Updated the JSON schema for taxonomy filters to ensure a consistent structure across front and back ends
* Updated filter_ids process to accept taxonomy filter file (necessary for proper staging)

### DIFF
--- a/docs/source/lib/EFI/Import/Filter/Taxonomy.pm.rst
+++ b/docs/source/lib/EFI/Import/Filter/Taxonomy.pm.rst
@@ -26,8 +26,8 @@ taxonomic classification.
 JSON FILE FORMAT
 ================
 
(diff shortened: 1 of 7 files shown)


rbdavid (approved):
This clarifies what structure the job_manager needs to handle. Thanks!

PR #315, opened 2026-04-29
Use map-reduce process to run the blastreduce step

### LINKED ISSUE #314: Address duckdb memory issues in blastreduce
The `blastreduce` step crashes on large jobs because duckdb can't process the input files and perform the reduce work in limited amounts of RAM.  Address the duckdb issue through the use of a map-reduce algorithm.

nilsoberg:
- Previously blastreduce logic was contained within one duckdb execution
- Resulted in memory issues
- This change performs blastreduce on each individual parquet (BLAST results) file
- Then merges and sorts all into one single 1.out.parquet file


pipelines/shared/blastreduce/render_reduce_sql_template.py
rbdavid, line 1: Should this be deleted if we are maintaining the `blastreduce_old` process block? This script is still needed for it.

pipelines/shared/templates/reduce-template.sql
rbdavid, line 1: Should this be deleted if we are maintaining the `blastreduce_old` process block? This template is still needed for that.

rbdavid (approved):
I think we should restore the deleted files if only so that `blastreduce_old` process block is still functioning. Otherwise, looks good.

PR #317, opened 2026-04-30
316 remove neighborhood connectivity dependency on gd

### LINKED ISSUE #316: Remove neighborhood connectivity dependency on GD
Remove the dependency on the Perl GD module from the neighborhood connectivity pipeline.  This module is not easy to install in some platforms (e.g. Docker).  This will entail a rewrite of the `make_color_ramp.pl` script.

nilsoberg:
Replaced the `get_connectivity.pl` and `make_color_ramp.pl` scripts with Python equivalents.  This necessitates updating the `pyEFI` library to add a shared color map function.


pipelines/neighborhoodconnectivity/color/make_color_ramp.py
```diff
@@ -0,0 +1,147 @@
...
+import os
+from pyEFI.nb_conn_colormap import get_nb_conn_colormap
+import sys
+
+def is_numeric(val):
+    try:
+        float(val)
+        return True
+    except ValueError:
+        return False
```
rbdavid: Is `val` a string and we want to see if its a string representation of a number? Or do we just want to see if `val` is a float or int already?

rbdavid: I guess this question would be cleared up with function annotations.

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
```diff
@@ -0,0 +1,147 @@
...
+                continue
+
+            parts = line.split('\t')
+            k = parts[0]
+            v = parts[1:]
+
+            # Check for the _META row to get the min max from that row, otherwise compute from data
+            if k == "_META" and len(v) >= 4:
+                min_val = float(v[1] if v[0] == "min" else v[3])
+                max_val = float(v[3] if v[2] == "max" else v[1])
```
rbdavid, line 72: Is this file format documented? This seems like some interesting logic.

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
```diff
@@ -0,0 +1,147 @@
...
+    Parameters
+    ----------
+        args
+            An argparse Namespace that contains the min and max that was obtained either from the
+            input file or command line arguments, and the output file path
+    """
+    # Setup the figure canvas
+    # The Perl script used 1650x150 pixels. 
+    # At 100 DPI, 16.5 x 1.5 inches provides the exact same canvas size.
+    fig, ax = plt.subplots(figsize=(16.5, 1.5), dpi=100)
```
rbdavid: DPI of 100 is somewhat low resolution. I think a minimum dpi of 300 is publication quality. That kinda breaks the simple math though.

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
```diff
@@ -0,0 +1,147 @@
...
+
+    if args.min is None or args.max is None or args.min == float('inf'):
+        print(f"Min/max error {args.min} {args.max}: saving empty image")
+        ax.axis('off')
+        plt.savefig(args.output_file, bbox_inches='tight')
+        return
```
rbdavid: Why would we ever want to do this?

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
```diff
@@ -0,0 +1,147 @@
...
+
+    print(f"Found color min {args.min} and max {args.max}")
+
+    # Set up the colormap and normalization range
+    if args.matplotlib_colormap:
+        # Use matplotlib standard
+        cmap = plt.get_cmap(args.matplotlib_colormap)
```
rbdavid: Should this be wrapped in a `try: except:` block in case args.matplotlib_colormap is incorrectly formatted or the specified map is not accessible? Just fall back to using the `get_nb_conn_colormap()` on exception thrown.

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
```diff
@@ -0,0 +1,249 @@
...
+import os
+from pyEFI.nb_conn_colormap import get_nb_conn_colormap
+import re
+import sys
+from typing import Any, Dict, List, Optional, Set, Tuple
+
+try:
+    import cdhit_reader
+except ImportError:
+    print("Warning: 'cdhit-reader' package is not installed. If you are using the --cdhit flag, please install it via 'pip install cdhit-reader'.", file=sys.stderr)
```
rbdavid: Is this check needed anymore if the python environment is built from the `requirements.txt`?

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
```diff
@@ -0,0 +1,249 @@
...
+
+    Parameters
+    ----------
+        line
+            Current line in the BLAST file that is being parsed
+        filter_ids
+            Set of cluster IDs in a cd-hit file
+        use_cdhit
+            true to search in the filter_ids set to return only if either query or subject are
+            in the set, false to return always
```
rbdavid: "false to return always" is strange wording...

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
```diff
@@ -0,0 +1,249 @@
...
+
+    Returns
+    -------
+        Tuple containing dictionary of degrees, and dictionary of neighbor list
+    """
+
+    filter_ids = set()
+    use_cdhit = False
+
+    if args.cdhit and os.path.isfile(args.cdhit):
```
rbdavid, line 120: If this evaluates to False, then `filter_ids` remains an empty set.

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
```diff
@@ -0,0 +1,249 @@
...
+
+    # Graph data structures
+    degree = collections.defaultdict(int)
+    neighbors = collections.defaultdict(list)
+
+    # Determine input type, either a BLAST file or a XGMML file, and create a callback to
+    # handle the parsing line
+    if args.input_blast:
+        in_file = args.input_blast
+        parse_fn = lambda l: parse_blast_line(l, filter_ids, use_cdhit)
```
rbdavid, line 131: Continuing the thought process from the line 125 comment, if `filter_ids` is an empty set, then calling this `parse_fn` will always return `(None, None)`. Is the `args.input_blast` file only intended to be used with a `args.cdhit` file that specifies cluster representatives?

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
```diff
@@ -0,0 +1,249 @@
...
+    filter_ids = set()
+    use_cdhit = False
+
+    if args.cdhit and os.path.isfile(args.cdhit):
+        filter_ids = get_cdhit_clusters(args.cdhit)
+        use_cdhit = True
+
+    # Graph data structures
+    degree = collections.defaultdict(int)
+    neighbors = collections.defaultdict(list)
```
rbdavid: "Degree" is referring to degree centrality, which is just the number of neighbors a node in a graph has. So, we really don't even need to track the degree dictionary at all.

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
```diff
@@ -0,0 +1,249 @@
...
+            Maximum neighborhood connectivity
+        NC
+            Neighborhood connectivity mapping
+    """
+    NC = {}
+    max_nc = 0
+    min_nc = float('inf')
+
+    # Calculate Neighborhood Connectivity
+    for nid, k in degree.items():
```
rbdavid: To continue along the thought process of only tracking neighbors (since degree is just the len of the neighbor's list). 

This could be:
```python
for node, node_neighbors in neighbors.items():
    node_degree = len(node_neighbors)
    nc_sum = sum([len(neighbors[id]) for id in node_neighbors])
```

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
```diff
@@ -0,0 +1,249 @@
...
+        nc_sum = sum(degree[n] for n in neighbors[nid])
+
+        # Perform rounding
+        val = int((nc_sum * 100) / k) / 100.0 if k > 0 else 0.0
```
rbdavid: What is the purpose of the rounding? Are we just trying to keep two decimal places? I'm not sure why we'd want to do that.

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
```diff
@@ -0,0 +1,249 @@
...
+        for nid in filter_ids:
+            if nid not in degree:
+                degree[nid] = 0
+                neighbors[nid] = []
+
+    return degree, neighbors
+
+def compute_connectivity(degree: Dict[str, int], neighbors: Dict[str, List[str]]) -> Tuple[float, float, Dict[str, Any]]:
+    """
+    Compute neighborhood connectivity and assign colors, and save a table with connectivity and colors.
```
rbdavid: This function does not do this description. No handling of colors is done here.

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
```diff
@@ -0,0 +1,249 @@
...
+
+    # Write output map
+    with open(args.output_map, 'w') as f:
+        # Save a header line containing min and max
+        if args.include_meta:
+            f.write(f"_META\tmin\t{min_nc}\tmax\t{max_nc}\n")
+
+        f.write("ID\tNC\tCOLOR\n")
+        for nid in sorted(NC.keys()):
+            f.write(f"{nid}\t{NC[nid]['nc']}\t{NC[nid]['color']}\n")
```
rbdavid: To continue the thought of rounding the `nc` values. Depending on the difference in min and max values for `nc`, decimal places beyond two points may be important to differentiate nodes. Instead of rounding the `nc` value before a color is computed for it, we can round the value here in the f-string handling. 
```
f.write(f"{nid}\t{NC[nid]['nc']:.2f}\t{NC[nid]['color']}\n")
```
where the `:.2f` tells the f-string to print the value as a float with two decimals.

This avoids prematurely rounding potentially important values while keeping the printed values "clean". Idk, just a suggestion.

rbdavid (approved):

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
nilsoberg: It's been "the way" for a while now and it's not documented.  I'm sure there are better ways of doing it and I would like to fix it.  However, in the interest of addressing the numerous other issues I think it's best to leave it the way it is.

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
nilsoberg: Yes

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
nilsoberg: No, hold over from testing when Docker wasn't updated

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
nilsoberg: Clarified

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
nilsoberg: That is acceptable because of the condition `if not use_cdhit` in `parse_blast_line`.

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
nilsoberg: If `use_cdhit` is False in `parse_blast_line`, then `filter_ids` is never checked and all IDs are returned, so that condition is acceptable.

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
nilsoberg: Correct

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
nilsoberg: John's request.

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
nilsoberg: Yes

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
nilsoberg: Moved this into the function body since it's only used once.

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
nilsoberg: To avoid crashing the Nextflow pipeline.  It could be argued that if there is no data that the pipeline should crash so I have changed this to a non-zero exit code.

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
nilsoberg: I changed this to 300 but that results in an image size of ~4500 pixels in width.

rbdavid: Yeah, we'd have to do the math to determine the correct size (in inches). To keep the perl code's pixel count: 

```
width = 1650 / 300
height= 150 / 300

fig, ax = plt.subplots(figsize=(width, height), dpi = 300)
```

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
rbdavid: Or can the absence of data be detected by the previous process block? Just a question. I think the non-zero exit code is a good idea.

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
rbdavid: I must have skimmed over that line. :+1:

pipelines/neighborhoodconnectivity/color/make_color_ramp.py
```diff
@@ -105,24 +100,23 @@ def plot_color_ramp(args: argparse.Namespace):
             input file or command line arguments, and the output file path
     """
     # Setup the figure canvas
-    # The Perl script used 1650x150 pixels. 
-    # At 100 DPI, 16.5 x 1.5 inches provides the exact same canvas size.
-    fig, ax = plt.subplots(figsize=(16.5, 1.5), dpi=100)
+    fig, ax = plt.subplots(figsize=(16.5, 1.5), dpi=300)
```
rbdavid, line 103: I responded in a comment that a quick calc can scale the figure size (units of inches) to achieve the higher DPI and the same number of pixels as the previous perl script created.

pipelines/neighborhoodconnectivity/connectivity/get_connectivity.py
```diff
@@ -184,15 +173,17 @@ def compute_connectivity(degree: Dict[str, int], neighbors: Dict[str, List[str]]
     min_nc = float('inf')
 
     # Calculate Neighborhood Connectivity
-    for nid, k in degree.items():
-        nc_sum = sum(degree[n] for n in neighbors[nid])
+    for nid, nb_list in neighbors.items():
+        k = len(nb_list)
+        nc_sum = sum(len(neighbors[n]) for n in nb_list)
```
rbdavid, line 178: We talked about this already, so I'm just leaving a note here to potentially come back to later. Since `nid` node is in the neighbor list for each node in `nb_list` are are effectively over-counting the `nc_sum` value. This may or may not be in accord with the publication that reported the neighborhood connectivity analysis. Something to look into.

rbdavid (approved):
We've already talked about these two comments, so everything looks great!

PR #323, opened 2026-05-07
319 optimize xgmml generation code

### LINKED ISSUE #319: Optimize XGMML generation code
All processes that write XGMML files are exceedingly slow on the HPC system.  Ensure that the performance is comparable to or better than the legacy code.

nilsoberg:
Three separate issues were addressed:

* When writing SSNs with high alignment scores, no edges were produced.  Due to a logic bug, this created an infinite loop.  This was fixed by checking for end of file from within the while loop in question.
* GNDs with large numbers of sequences were not able to be written by the tool due to a missing parameter.  Smaller datasets were processed successfully because the code in question was used to optimize writing of large datasets.
* Writing large SSNs caused jobs to crash because of excessive memory usage.  This was caused by loading all edges into memory from the blast results file (`2.out`) and then passing it to the EFI::SSN::XgmmlWriter class as a data structure.  This problem is solved by passing a "generator" to the EFI::SSN::XgmmlWriter class that read lines one at a time from the file, avoiding loading the entire file into memory.

### DIFF
--- a/conf/default_params.config
+++ b/conf/default_params.config
@@ -24,6 +24,7 @@ params {
     threshold_metric = "alignment_score"
     min_length = 0
     max_length = 65000
+    max_ssn_edges = 200000000
 
     // GNT/GND parameters
     nb_size = 20
--- a/lib/EFI/GNT/GND/Util.pm
+++ b/lib/EFI/GNT/GND/Util.pm
@@ -13,7 +13,7 @@ use EFI::Util::Colors;
 
 sub new {
     my $class = shift;
-    my %args;
+    my %args = @_;
 
     my $self = {};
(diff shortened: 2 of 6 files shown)


rbdavid (approved):
Have we set a reasonable default value for the max_edges parameter, considering John's size of jobs.

nilsoberg:
> Have we set a reasonable default value for the max_edges parameter, considering John's size of jobs.

I 1) updated the default in the perl script to `200000000`, and 2) added a parameter in the `default_params.config` file (`max_ssn_edges`) to control that value.  In the future the job manager and front end could be adapted to make this an "administrator" or "dev site" setting.  For now, that's not necessary.

PR #324, opened 2026-05-08
Apply fraction filter to proper sequences

### LINKED ISSUE #156: Keep SwissProt sequences when applying fraction filter
When applying the fraction filter to retain only a portion of the sequences, ensure that any sequences with SwissProt annotations are retained.

### LINKED ISSUE #166: Apply the fraction filter to families only
Once merged, PR #151 implements the fraction filter. If set, this filter is applied to all sequences input to the workflow, not just those PFAM/InterPro family/clan IDs input via the Families path (branch B) or "Protein Family Addition Options". The latter is the intended usage. 
- [ ] Develop the `EFI::Import::Filter::Fraction` module to check the "source" sequence metadata/attribute for "FAMILY" (or equivalent), loop over each unique family/clan ID, and apply the fraction filter on that subset of sequences.

### LINKED ISSUE #320: Family addition fraction option doesn't work correctly
For Option A (and possibly others), when adding a family and selecting a fraction, the wrong fraction is used.

nilsoberg:
- Only sequences coming from a family are filtered.  This means that Accession, BLAST, FASTA sequences are not filtered out when applying the fraction filter.
- Do not filter out SwissProt sequences when applying fraction filter.

Closes #156, closes #166, closes #320


nilsoberg:
Also deleted `EST::Metadata` module; hasn't been used in over a year.

rbdavid (approved):

PR #326, opened 2026-05-14
178 add sequences to ssn nodes for fasta input files

### LINKED ISSUE #178: Add sequences to SSN nodes for FASTA input files
Add the FASTA sequence to the SSN node for the unknown sequence ID types for FASTA input file job types.

nilsoberg:
This PR contains the following changes:

* Insert FASTA sequences for Option C into SSN
* Insert BLAST input sequence into SSN
* Improve naming of unknown sequences (prefix with `ZZ` instead of single `Z` -- avoid clashes with UniProt database)
* Lengthen unknown sequences from 6 to 10 to allow more sequences
* Rename BLAST input sequence from `ZINPUTSEQ` to `ZZINPUTSEQ` for consistency with unknown sequences
* Fix uninitialized value bug in writing sequence metadata (`EFI::Sequence::Collection`)


rbdavid (approved):
Do we have test module(s) that check for correct behavior for these updates? Especially for the unknown sequence as input in the Fasta or Blast input modes. A quick glance at the testing data set, doesn't seem like we have any input files with unknown or no header information.

nilsoberg:
In the most recent commits, I refactored the shared `get_source_ids` process into a workflow.  I moved the logic that was contained in the process into a subworkflow (`GET_SOURCE_IDS`).  This change allows us to create an input file channel from the `params.input_file` rather than embedding it as a variable in the individual process blocks.  We can now leverage Nextflow's staging capability to use any file as the input rather than restricting it to certain path subset.  This is vital for Docker.

rbdavid (approved):
I think this all looks right. Good addition for the test module!

PR #329, opened 2026-05-21
327 remove self alignments after restore condensed

rbdavid:
This PR is a large rework of the `restore_condensed_sequences.py` script to use the duckdb python api instead of doing the work using raw python code. First the cd-hit cluster file is parsed into a parquet file. Then, the cluster parquet and the alignment parquet file are used in one query to expand the cluster alignment results to full sequence set. A final blast alignment file (default name: 1.out.parquet) is written that contains the expanded blast results without any "self-edges" or self-alignment results (which were present previously). 

The old process block and python code are kept, for now, just with slight name change.

The above change should solve the Convergence Ratio errors we've been seeing where there were N-too-many edges (where N is the number of sequences in the SSN cluster), arising from self-alignment results. This also implements a bug fix relative to the production site, where intra-CD-HIT-cluster alignment results are written when they previously were not. There is a CLI argument to switch functionality between old and new behavior for validation/testing purposes.

Additionally, the nextflow `publishDir` lines for "multiplex" calls now also confirm that the `params.sequence_version` is set to "uniprot" (which is set by default in `default_params.config`) before any CD-HIT condensing and uncondensing is done. UniRef sequence sets will never need the CD-HIT call to be performed, so that will save us some time.

### DIFF
--- a/pipelines/shared/condense/restore_condensed_sequences.py
+++ b/pipelines/shared/condense/restore_condensed_sequences.py
@@ -1,5 +1,6 @@
 
 import argparse
+import os
 import re
 
 import duckdb
@@ -22,11 +23,13 @@ def get_args() -> argparse.ArgumentParser:
     parser.add_argument(
         "--cd-hit-cluster",
         type=str,
+        required=True,
         help="Path to the CD-HIT cluster file"
     )
     parser.add_argument(
         "--condensed-blast",
         type=str,
+        required=True,
         help="Path to the condensed BLAST result parquet"
     )
     parser.add_argument(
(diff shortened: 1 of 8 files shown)


nilsoberg (approved):

pipelines/shared/condense/restore_condensed_sequences.py
```diff
@@ -1,105 +1,277 @@
...
+    parser.add_argument(
+        "--duckdb-temp-dir",
+        type=str,
+        default="./duckdb",
+        help="Location DuckDB should use for temporary files"
+    )
+    parser.add_argument(
+        "--use-old-method",
+        type=bool,
+        default=True,
```
rbdavid, line 69: This should be changed to `default=False` once validation tests with the EFI Tools v1 are complete. The "old method" is a source of important edges missing in SSN files.

PR #334, opened 2026-05-28
333 only include swissprots in swissprot table

### LINKED ISSUE #333: Only include SwissProts in SwissProt table
The Color SSN and Cluster Analysis utilities provide a table that lists SwissProt IDs and the description.  Currently this file erroneously lists all IDs in the input.  The output should only include IDs that are SwissProt-reviewed.

nilsoberg:


rbdavid (approved):
Solves two problems.

PR #336, opened 2026-05-28
(HOTFIX) restore_condensed_sequences.py CD-HIT parsing

### LINKED ISSUE #335: CD-HIT parsing needs to be reworked
https://github.com/EnzymeFunctionInitiative/EST/blob/35f7ca208b088519a87f529e5bf714cc18a966b3/pipelines/shared/condense/restore_condensed_sequences.py#L147-L161

CD-HIT ordering of clusters is not rep_seq, child_seq, child_seq, etc. It can be out of order. The current code will be bugged when that is the case.

rbdavid:
Fixes the cluster file parsing when CD-HIT outputs a cluster's rep sequence in the non-zero index of the cluster member list. Previous version assumed that the rep sequence would always be index 0 of the list; that is not necessarily the case.

Closes #335

### DIFF
--- a/pipelines/shared/condense/restore_condensed_sequences.py
+++ b/pipelines/shared/condense/restore_condensed_sequences.py
@@ -129,20 +129,22 @@ def process_cd_hit_clusters(clstr_file: str, output_path: str):
 
     Creates the map file in f"{output_path}/map.parquet"
     """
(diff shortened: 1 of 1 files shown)


nilsoberg (approved):
Looks good.

PR #338, opened 2026-05-29
337 fix statsjson for ssn utilties

### LINKED ISSUE #337: Fix stats.json for SSN utilties
The neighborhood connectivity pipeline doesn't correctly output a `stats.json`; it should contain the `file_stats` key that stores the file stats such as size, but this is currently stored at the top level.  Additionally, convergence ratio pipeline should output a `stats.json`.  It currently does not.

nilsoberg:
Properly output `stats.json` files from Neighborhood Connectivity, Convergence Ratio, and Generate SSN pipelines.  Also correct syntax errors in a python script.

### DIFF
--- a/pipelines/convergenceratio/convergenceratio.nf
+++ b/pipelines/convergenceratio/convergenceratio.nf
@@ -20,14 +20,15 @@ process compute_blast_conv_ratio {
 }
 
 process merge_conv_ratios {
-    publishDir params.final_output_dir, mode: "copy", pattern: "{conv_ratio.tab}"
+    publishDir params.final_output_dir, mode: "copy", pattern: "{conv_ratio.tab,stats.json}"
 
     input:
         path stats_files
         path cr_table
 
     output:
-        path "conv_ratio.tab"
+        path "conv_ratio.tab", emit: "conv_ratio"
+        path "stats.json", emit: "stats"
 
     script:
     """
@@ -35,6 +36,7 @@ process merge_conv_ratios {
         --ssn-conv-ratio ${cr_table} \
         --stats ${stats_files} \
         --output conv_ratio.tab
+    echo "{}" > stats.json
     """
 }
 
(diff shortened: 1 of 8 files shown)


pipelines/convergenceratio/convergenceratio.nf
```diff
@@ -20,21 +20,23 @@ process compute_blast_conv_ratio {
...
+        path "stats.json", emit: "stats"
 
     script:
     """
     python $projectDir/statistics/merge_conv_ratios.py \
         --ssn-conv-ratio ${cr_table} \
         --stats ${stats_files} \
         --output conv_ratio.tab
+    echo "{}" > stats.json
     """
```
rbdavid, line 40: Just to clarify my dms about the convergence ratio job not having a stats.json file. If the job doesn't have any stats that would go inside that file, then it doesn't need to be created. The job_manager is not dependent on the presence of that file. But, if results need to be parsed from a different (non- stats.json) file, that's when I need to know about it.

rbdavid (approved):

rbdavid:
I've run a large families job with a fragment filter applied. Its stats.json file contents are:
```
{
    "num_ids": 6648161,
    "num_filter_fragment": 38641,
    "num_uniref50_in_family": 829481,
    "num_uniref90_in_family": 830118,
    "num_full_family": 830297,
    "convergence_ratio": 0.040840535733504435,
    "num_blast_edges": 113558289,
    "num_unique_ids": 74573
}
```

I'm very confused how `num_full_family`, `num_uniref50_in_family`, and `num_uniref90_in_family` are nearly equivalent. I know we don't use these numbers but what's going on there? 

I'm also a bit confused how `num_full_family` and `num_ids` are different. This job used 20+ pfam ids and uniref50 as the sequence source. Is `num_ids` counting all sequences associated with each family from the UniProt level, whether or not a sequence might be in multiple families? And then `num_full_family` is the unique set of sequence IDs?

rbdavid:
My comment above is out-of-date with the changes made in this PR. Ignore me. :)

Issue #301, opened 2026-04-17
Update job scheduler specific configs


rbdavid:
For both PBSPro and Slurm config files, update the process configuration blocks to use cluster options and better handle allocation management.

rbdavid:
Also, add `label` and `tag` fields for relevant process blocks to more cleanly define allocation usage for those specific types of processes.

rbdavid:
This will be closed when #347 closes.

rbdavid:
New development to control resource allocations for process blocks means this issue can be closed.

Issue #191, opened 2025-07-28
Note: `clusterOptions '-V'` may be required


rbdavid:
Depending on how jobs are submitted to cluster-managing executors (pbspro, slurm, etc), it may be required to add `clusterOptions '-V'` to pass environment variables to process blocks that get submitted as independent jobs on the queue. For PBS Pro, the `'-V'` parameter flags that all environment variables from the submitting environment get passed to the child process that gets submitted. The analogous sbatch command argument is `'--export=All'`. 

If required, changes needs to happen within the nextflow pipeline codes (not in the config files); e.g. the all-by-all blast block in EST needs to have a `clusterOptions '-V'` line added in the process' preamble space. Currently, this will be left as a "Note" issue to remind ourselves what may be required in certain instances. It is an executor-specific change after all.

rbdavid:
Documentation for the latest nextflow version (v25.04.6) show the `clusterOptions` configuration argument being used in the nextflow config file blocks. Maybe changes to the core nextflow scripts isn't necessary.

nilsoberg:
How was `clusterOptions -V` being used on the recent PBS Pro implementation?  Did we manually edit the `.nf` file for the deliverable or were we able to put it in the `pbspro.config` file?

rbdavid:
We manually edited the nextflow pipeline scripts.

I tried to add the `clusterOptions '-V'` in the config files but didn't try to debug when those attempts failed. Documentation for nextflow v24.04 is difficult to find and traverse (in regards to PBSPro executor, relevant docs are https://github.com/nextflow-io/nextflow/blob/STABLE-24.04.x/docs/executor.md#pbs-pro and https://github.com/nextflow-io/nextflow/blob/STABLE-24.04.x/docs/process.md#array). The array documentation makes it seem that the clusterOptions is included in the process block. But there's other documentation pages that make it seem like the config file can hold the `clusterOptions` info (https://github.com/nextflow-io/nextflow/blob/STABLE-24.04.x/docs/config.md#scope-process).

PR #339, opened 2026-06-01
182 - Sort fasta before axa step

### LINKED ISSUE #182: Ensure BLAST results are consistent
Results from the EST pipeline can produce different results depending on the order of sequences output by the `cat_fasta_files` pipeline step.  Different sequence pairs can have identical scores.  This likely caused by an issue in the `blastreduce` pipeline step, which always takes a specific pair of sequences (likely the first one).  One solution is to order sequences during the `cat_fasta_files` step or in `pipelines/est/templates/reduce-template.sql`.

Update the pipeline to consistently return results no matter the order of the input sequences.

A simple sort can be added in `cat_fasta_files`, however alternatives should be examined to determine the best solution.

```
diff --git a/pipelines/est/est.nf b/pipelines/est/est.nf
index 9274ee3..8d150ad 100644
--- a/pipelines/est/est.nf
+++ b/pipelines/est/est.nf
@@ -102,9 +102,15 @@ process cat_fasta_files {
         """
         $cat_cmd
         perl $projectDir/import/append_blast_query.pl --blast-query-file ${params.blast_query_file} --output-sequence-file all_sequences.fasta
+        awk '/^>/{key=\$1} {print key, \$0}' all_sequences.fasta | sort -k1,1 -s | cut -d' ' -f2- > sorted.fasta
+        mv sorted.fasta all_sequences.fasta
         """
     } else {
-        cat_cmd
+        """
+        $cat_cmd
+        awk '/^>/{key=\$1} {print key, \$0}' all_sequences.fasta | sort -k1,1 -s | cut -d' ' -f2- > sorted.fasta
+        mv sorted.fasta all_sequences.fasta
+        """
     }
 }
```

rbdavid:
Past work on duckdb analyses of blast output (w/in `all_by_all_blast()`, `blastreduce()`, and `restore_condensed()` process blocks) should standardize the lexicographical sorting, which should solve the concerns of the original post of #182. But there's additional sorting we can do to ensure a standardized creation of the input fasta shards to the `all_by_all_blast()` process. This change standardizes the organization of those fasta files _and_ does a load-balancing of the sequences spread across those shards. This load balancing should result in more even distribution of computationally expensive (long) sequences and so avoid instances where a random set of `all_by_all_blast()` instances take much longer than the rest.

Closes #182


nilsoberg (approved):
It looks good.  The only thing I would check -- assuming you haven't yet -- is that the seqkit call doesn't run out of memory with large sequence sets.

rbdavid:
> It looks good. The only thing I would check -- assuming you haven't yet -- is that the seqkit call doesn't run out of memory with large sequence sets.

That kinda bleeds into what I was envisioning was my next _major_ task, which was to revamp the nextflow configuration files to better control how certain memory/compute hungry process blocks are handled and resources are allocated. As seqkit notes in their documentation, `sort` and `split2` can be memory hungry (I use `--two-pass` arg to try to minimize this). Also, seqkit has a `--threads` argument (set to 4 by default) so we need to explicitly control that argument. All-in-all, any process block that uses seqkit should probably be given its own process executor with extra allocation for memory and threads.

rbdavid:
That above discussion is related, in part, to #301 and #191.

rbdavid:
Added a process block call to remove self-alignment rows when those are not used by jobs that multiplex/condense and then run the `restore_condensed` process block. Any UniRef source-db job currently skips the multiplex/condense step and so needs to run through this new process block to remove the self-alignment rows before "publishing" 1.out.parquet.

nilsoberg (approved):

PR #353, opened 2026-06-10
Add support for UniRef and domain length histograms

### LINKED ISSUE #352: Compute UniRef length histogram
The est pipeline creates histograms for UniProt sequences but not for UniRef sequences when UniRef is used.  Update the pipeline to also create a length histogram for the UniRef sequences.

nilsoberg:
- Refactor length histogram code to support length mapping file in addition to FASTA sequence file
- Support creation of a length mapping file
- Histogram computation process blocks are no longer shared (moved out of reporting.nf)
- Move shared/python/visualization scripts to shared/visualization to follow pattern
- Add new script for retrieving length data for UniProt and domains
- Addition of DBD::DuckDB module to Perl to support parquet reading
- Fix Nextflow channel consumption issue when passing channels in est.nf and taxonomy.nf


pipelines/est/subworkflows/reporting.nf
```diff
@@ -55,28 +61,94 @@ process visualize_boxplot_stats {
...
+    """
+    perl $projectDir/visualization/retrieve_sequence_lengths.pl \
+        --fasta-lengths-parquet ${fasta_lengths_parquet} \
+        --fasta-lengths fasta_lengths.tab \
+        --accession-table ${accession_table} \
+        --sequence-metadata ${sequence_metadata} \
+        --uniprot-lengths uniprot_lengths.tab \
+        --config ${params.efi_config} \
+        --db-name ${params.efi_db}
+    """
```
rbdavid, line 116: This should be a seqkit call instead of a whole perl script call. The sequences are stashed in the `all_sequences.fasta` file, which is created after domain regions are gathered, so the sequences in that fasta file are of the correct length.
Something like:
`seqkit fx2tab all_sequences.fasta --length --name --no-qual --only-id` will print: `SEQID\tseq_len\n` on each line.

pipelines/est/subworkflows/reporting.nf
```diff
@@ -55,28 +61,94 @@ process visualize_boxplot_stats {
...
         fasta_file
         accession_table
         import_stats
+        sequence_metadata
 
     main:
         seq_versions = ['uniprot']
         if (params.sequence_version == "uniref50") {
-            seq_versions = ["uniprot", "uniref90", "uniref50"]
+            seq_versions = ["uniprot", "uniref50"]
```
rbdavid, line 131: Why are we not doing the uniref90 expansion for `sequence_version == "uniref50"`?

pipelines/est/subworkflows/reporting.nf
rbdavid: Talked in person about this. My interpretation is not accurate since we need to gather uniref90 and uniprot sequence IDs and lengths for "expanding" the sequence set for visualization purposes. We need to query the database to find sequence IDs that are tiers above the currently used `sequence_version` (UniRef50 or UniRef90).

rbdavid: Keeping this comment thread for documentation purposes, if that makes sense.

pipelines/est/visualization/retrieve_sequence_lengths.pl
```diff
@@ -0,0 +1,126 @@
...
+
+
+my $ids = new EFI::Sequence::Collection();
+$ids->load($opts->{sequence_metadata}, $opts->{accession_table});
+
+my @tableIds = $ids->getAllSequenceIds();
+
+my $fastaLengths = {};
+if ($opts->{fasta_lengths_parquet}) {
+    $fastaLengths = parseFastaLengths($opts->{fasta_lengths_parquet});
```
rbdavid, line 44: Why is this an optional input argument?

pipelines/shared/visualization/compute_length_histogram.py
```diff
@@ -109,35 +109,89 @@ def compute_sequence_lengths(fasta_file: str, valid_ids: Set[str] = None) -> Lis
...
+                if valid_ids == None or parts[0] in valid_ids:
+                    sequence_lengths.append(parts[1])
+
+    except FileNotFoundError:
+        print(f"Error: length mapping file not found at '{length_mapping_file}'", file=sys.stderr)
+        sys.exit(1)
+
+    return sequence_lengths
+
+def get_sequence_lengths(accession_table: str, fasta_file: str, length_mapping_file: str, seq_type: str = None) -> List[int]:
```
rbdavid, line 142: Function formatting and annotations are a bit misleading. All three of these files are required inputs and they're being used as positional arguments, so if, for example, the input for the `accession_table` is `None`, we still need to declare `None` in the zeroth element of the function call. I think this can be cleaned up and made more correct: I think the option to declare arguments via position should be revoked and the default values for these arguments should be set to `None`.

```
def get_sequence_lengths(*, accession_table: str | None = None, fasta_file: str | None = None, length_mapping_file: str | None = None, seq_type: str | None = None) -> List[int]:
```
where the `*` symbol in the line prevents the usage of positional arguments. The user would then need to call this func as:

```
list_of_lens = get_sequence_lengths(
    accession_table = args.accession_table,
    fasta_file = args.fasta_file,
    length_mapping_file = args.length_mapping_file,
    seq_type = args.seq_type
)
```
This completely removes ambiguity from the usage lines _and_ the annotation information will be correct as well. Note, this is purely a clean-up and formatting suggestion. No priority to update if you  don't wish to.

pipelines/shared/visualization/compute_length_histogram.py
```diff
@@ -109,35 +109,89 @@ def compute_sequence_lengths(fasta_file: str, valid_ids: Set[str] = None) -> Lis
...
     if accession_table and seq_type:
         valid_ids = parse_accession_table(accession_table, seq_type)
     else:
         valid_ids = None
 
-    sequence_lengths = compute_sequence_lengths(fasta_file, valid_ids)
+    if fasta_file:
+        sequence_lengths = compute_sequence_lengths(fasta_file, valid_ids)
+    elif length_mapping_file:
+        sequence_lengths = load_length_mapping(length_mapping_file, valid_ids)
```
rbdavid, line 173: Are there potential instances where both a `fasta_file` and `length_mapping_file` are passed to the func?

rbdavid (approved):
Mainly leaving notes for future consideration. Like using seqkit when possible to gather sequence lengths. Duckdb interface in perl now as well as python and cli calls. Ideally we only used one of those interfaces in a standard fashion. But all in all, I think this is ready to be merged.

pipelines/est/visualization/retrieve_sequence_lengths.pl
nilsoberg: This isn't (or doesn't need to be) used when running Option B, for example.

pipelines/est/subworkflows/reporting.nf
nilsoberg: When the user selects UniRef50, we don't provide length histograms for UniRef90.  I believe this had something to do with UniRef90 membership possibly crossing family boundaries.  The legacy site only provides one UniRef (the primary selection) and UniProt.

pipelines/shared/visualization/compute_length_histogram.py
nilsoberg: No -- `fasta_file` is used in the cluster analysis pipeline.  `length_mapping_file` is used by est.

pipelines/shared/visualization/compute_length_histogram.py
nilsoberg: Thanks for the feedback and I see the point.  In the interest of meeting deadlines I'm going to leave it as-is.

PR #355, opened 2026-06-12
Fix nested SSN and UniRef SSN attributes

### LINKED ISSUE #159: Use custom character for annotation row separators
Allow characters other than ^ to be used to separate metadata when using metanodes (e.g. UniRef, RepNodes).  Changes made in PR #151 should be included in this issue.

              As in a comment I left for PR #151, a multi-character separator could be used to avoid potential mis-splitting annotation strings that contain the ^ character.

_Originally posted by @rbdavid in https://github.com/EnzymeFunctionInitiative/EST/pull/155#discussion_r2044828832_

### LINKED ISSUE #343: Fix nesting behavior with SSN utility pipelines
When coloring an SSN (e.g. job 644), the node attributes are improperly nested when rewriting the SSN.

```
<node id="A0A017HT19" label="A0A017HT19">
<att name="Sequence Source" type="list">
<att name="Sequence Source" value="FAMILY" type="string" />
<att name="Sequence Count Cluster Number" type="integer" value="5" />
<att name="Node Count Cluster Number" type="integer" value="5" />
<att name="Cluster Sequence Count" type="integer" value="12" />
<att name="Cluster Node Count" type="integer" value="12" />
<att name="node.fillColor" type="string" value="#FF00FF" />
<att name="Node Count Fill Color" type="string" value="#FF00FF" />
</att>
```

### LINKED ISSUE #346: SSNs don't include UniRef cluster information
"UniRef Cluster Size and UniRef Cluster IDs node attributes are provided in the UniRef SSNs. But, the cluster sizes are 1 and the cluster IDs are the node IDs."

nilsoberg:
This PR resolves several issues:

1. UniRef cluster information and other injected attributes were being incorrectly positioned in SSN XGMML files when encountering nested <att type="list"> elements.  This error occurs when using repnode networks as input to pipelines that rewrite SSNs.
2. UniRef cluster attributes were not being properly inserted into SSNs
3. Missing import stat

To fix this, the following changes were made:

1. Added nested tag tracking to solve issues with 'list' type tags when using repnode networks as input.  Closes #343
2. Add UniRef cluster attributes by ensuring node attribute values are expanded.  Closes #346
3. The number of UniRef clusters in a family is now added.  This attribute was previously not being tracked and will now be provided to the user
4. Simplified XGMML attribute writer by removing flag indicating if new attributes should be prepended or appended (default behavior was appended and the flag wasn't used)
5. Fixed error with multiple values in sequence metadata.  Previously the delimiter was a regular expression special character (`^`) and wasn't being properly escaped.  In order to create a robust format, the code is now using the ASCII "unit separator" character which will avoid any clashes with node attributes that have `^` as a value.  Additionally, this is now hardcoded (through the use of a constant) rather than allowing consumers of the `EFI::Sequence` class to specify the delimiter.  Code that was previously manually splitting (to work around a logic error when using `split`--also fixed) now relies exclusively on the `EFI::Sequence::unpackAttributeValue` function to get lists of attributes.  The file is only read by the machine, so the user doesn't need to edit metadata file and won't encounter the new character.  Closes #159


lib/EFI/Import/Source/Family.pm
```diff
@@ -78,11 +78,12 @@ sub loadFromSource {
...
 
     # Add the UniRef IDs to the metadata file
-    $self->addUnirefIds($destSeqData, $self->{sequence_version}, $uniprotUniref);
+    my $numUnirefIds = $self->addUnirefIds($destSeqData, $self->{sequence_version}, $uniprotUniref);
     my $numFromFamily = 0;
 
     if ($self->{sequence_version} ne SEQ_UNIPROT) {
-        $self->addStatsValue("num_full_family", $numIds);
+        $self->addStatsValue("num_full_family", $numIds); # The number of UniProt IDs in the families
+        $self->addStatsValue("num_family", $numUnirefIds); # If this is a UniRef job, this is the number of UniRef clusters in the families
```
rbdavid, line 86: Is this dependent on the UniRef sequence_version set by user? Or should each job get a value for UniRef90 and UniRef50 clusters in the families?

lib/EFI/Import/Source.pm
```diff
@@ -100,15 +100,24 @@ sub addStatsValue {
...
     my $self = shift;
     my $seqData = shift;
     my $seqVersion = shift || $self->{sequence_version};
     my $unirefIds = shift;
 
     if ($unirefIds) {
-        map { $seqData->associateUnirefIds($_, $unirefIds->{$_}->[0] || "", $unirefIds->{$_}->[1] || ""); } keys %$unirefIds;
-        return;
+        my %uniqueIds;
+        my $unirefIndex = $seqVersion eq SEQ_UNIREF90 ? 0 : 1;
```
rbdavid, line 114: What about UniRef50? Maybe I'm not following why we're testing for seqVersion UniRef90 only.

lib/EFI/Sequence.pm
```diff
@@ -43,11 +43,12 @@ sub getId {
 # public
 sub getAttribute {
     my $self = shift;
-    my $attr = shift || die "Require attribute name";
+    my $attr = shift or die;
```
rbdavid: Should we keep the error message so that we know why the `getAttribute()` call failed?

rbdavid (approved):
There were a lot of fixes being implemented in this PR. I think I kept all of them in mind while reviewing. Looks good, I think.

lib/EFI/Import/Source/Family.pm
nilsoberg: It depends on the input.  Due to the possibility that UniRef90 IDs in a UnIRef50 cluster could be members of a different family, we only use/report a single UniRefXX.

lib/EFI/Import/Source.pm
nilsoberg: If the user uses UniProt, this value doesn't get saved.  Logic in the function that calls this determines if this value gets saved.  In that case the value computed in this particular line is inconsequential.

lib/EFI/Sequence.pm
nilsoberg: This should never occur but yes, it can be kept.

PR #363, opened 2026-06-15
362 add domain data to ssns

### LINKED ISSUE #362: Add domain data to SSNs
Currently domain SSNs are generated without domain regions.  Update the generatessn pipeline so that the fasta file is used to determine the nodes; this will contain the domain data.  Use the sequence metadata files to obtain the node attributes.  Then continue as normal to use the 2.out file to obtain the edges.

nilsoberg:
Support domain sequence IDs in the SSN output, full and repnode.

- Use FASTA file as the source of the IDs, not the node metadata file
- Use shared metadata from the UniProt ID for each domain ID that shares the UniProt ID
- Edges isn't safe to get node IDs from, because it will ignore singletons
- Also optimize `strip_domain` function since it will be called many times

### DIFF
--- a/lib/EFI/SSN/XgmmlWriter.pm
+++ b/lib/EFI/SSN/XgmmlWriter.pm
@@ -10,7 +10,7 @@ use lib dirname(abs_path(__FILE__)) . "/../..";
 
 use EFI::Annotations;
 use EFI::Annotations::Fields qw(:annotations :source FIELD_CYTOSCAPE_COLOR);
-use EFI::Sequence::Type qw(is_unknown_sequence SEQ_FULL SEQ_DOMAIN);
+use EFI::Sequence::Type qw(is_unknown_sequence strip_domain SEQ_FULL SEQ_DOMAIN);
 
 use parent qw(EFI::Xgmml::Writer);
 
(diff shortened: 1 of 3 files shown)


lib/EFI/Sequence/Type.pm
```diff
@@ -54,8 +54,8 @@ sub get_sequence_type {
 
 sub strip_domain {
     my $id = shift;
-    $id =~ s/^([^\:]+):\d+:\d+$/$1/;
-    return $id;
+    my $colonPos = index($id, ":");
```
rbdavid, line 57: Is there ever an instance where `":"` exists in an `$id`? Like, if the user uploads a fasta file with weird header lines, do those header lines get removed/converted to some "ZZZ"-prefixed string? I'm trying to think of corner cases where user-input can really mess things up.

lib/EFI/SSN/XgmmlWriter.pm
```diff
@@ -325,14 +325,16 @@ sub makeNodeAttributes {
     my $id = shift;
     my $fields = shift;
 
+    my $uniprotId = strip_domain($id);
```
rbdavid, line 328: See the above question about corner cases, because this is called on every node.

rbdavid (approved):
I have some further questions that are a bit convoluted and probably low-importance. So, I think this PR solves the big problem of domain attributes missing from SSNs. Approved.

PR #367, opened 2026-06-16
366 - Fix "eukaryota_no_fungi" predef taxonomy filter

### LINKED ISSUE #366: Fix predef taxonomy filter eukaryota_no_fungi
Image

Image

The predefined taxonomy filter for "eukaryota_no_fungi" is not being applied correctly.

rbdavid:
Fixes the json formatting of the taxonomy filter to use the "negate" keyword instead of the "operator" keyword for the "condition" elements. Still needs to be tested.

Closes #366

### DIFF
--- a/pipelines/shared/assets/predefined_taxonomy_filters.yml
+++ b/pipelines/shared/assets/predefined_taxonomy_filters.yml
@@ -62,27 +62,27 @@
             {
                 "field": "phylum",
                 "value": "Ascomycota",
-                "operator": "NOT"
+                "negate": "true"
             },
             {
                 "field": "phylum",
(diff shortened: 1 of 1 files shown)


rbdavid:
Relevant code: 
https://github.com/EnzymeFunctionInitiative/EST/blob/287750c8bbff88f7f05e0a8fce63a31763dcb773/lib/EFI/Import/Filter/Taxonomy.pm#L108-L123

So elements in the "condition" array may or may not have keywords "negate", "exact", "field", and "value". The "operator" keyword is only used for the outer filter json object.

nilsoberg (approved):

PR #375, opened 2026-06-18
372 - Repnode associated process block configuration

rbdavid:
- [x] Each repnode has a cd-hit call, which has a hardcoded memory allocation (set by `-M` parameter). The associated process blocks for cd-hit calls generally used the standard `"local"` executor, which did or did not have a predefined memory allocation. This resulted in the cd-hit expected memory allocation to be out of sync with the available memory allocation given to the process block. Instead, the cd-hit process blocks have been given a label `APP_cd_hit`, which tells them to use the scheduler executor (when relevant) and explicitly defines a memory and thread allocation for processes with that label.
- [x] The `TASK_create_ssn` label has been added. When relevant, this label ensures that the `create_repnode_ssn` process block is performed 11 times in perfectly parallel strategy rather than being done serially on the `"local"` executor.
- [x] ~~An element of the XGMML writer seems to be a slow down. Find that bottleneck and make it more efficient.~~ See #376 for further investigation into this software bottleneck. This fix is beyond the scope of the current PR.

There's some interesting development that can happen with the config file work started here. ~~Nextflow handling of parameters is done \_before\_ process block configuration is cemented and applied during the workflow, so if we wanted to dynamically handle resource allocations for memory/cpu-hungry process blocks, we could do so in various ways and at various levels.~~ Edit: I'm not sure this is true. I need to at least call the `includeConfig` call before `process { ... }` configuration block in the central configuration file, otherwise nextflow doesn't know what `params.cd_hit_memory` is (for example). 

For example, if we wanted jobs to run on high memory nodes, the whole job doesn't necessarily need to run on that high-mem compute resource. Instead, we can identify the process blocks that require high memory and set the process' `"queue"` to be "efi-mem" or equivalent on other implementations. This could enable us to process large SSNs and avoid any sort of hard-cutoff for input data. Edit: Considering the above edit, this statement should still be possible because we can overwrite the parameters' values from those in the `default_params.config` to ones written explicitly in the `params.json` file.

### DIFF
--- a/conf/docker.config
+++ b/conf/docker.config
@@ -22,6 +22,11 @@ process {
         memory = params.cdhit_memory_limit
     }
 
+    withLabel: 'TASK_create_ssn' {
+        cpus = 1
+        memory = '20 GB' // is this enough?
+    }
+
 
 
 
--- a/conf/docker.config
+++ b/conf/docker.config
@@ -6,6 +6,11 @@ nextflow.enable.configProcessNamesValidation = false
 
 docker.enabled = true
 
+/*
+Include a config file containing the default values for workflows
+*/
+includeConfig './default_params.config'
+
 process {
     cpus = 1
     memory = 5.GB
@@ -19,11 +24,8 @@ process {
 
     withLabel: 'TASK_create_ssn' {
         cpus = 1
-        memory = '20 GB' // is this enough?
+        memory = '20 GB'
     }
-
-
-
 
     withName: blastreduce {
         memory = 20.GB
@@ -41,9 +43,4 @@ process {
         cpus = 1
     }
 }
-
-/*
-Include a config file containing the default values for workflows
-*/ 
-includeConfig './default_params.config'
 
--- a/conf/docker.config
+++ b/conf/docker.config
@@ -24,7 +24,7 @@ process {
 
     withLabel: 'TASK_create_ssn' {
         cpus = 1
-        memory = '20 GB'
+        memory = 20.GB
     }
 
     withName: blastreduce {
--- a/conf/docker.config
+++ b/conf/docker.config
@@ -6,11 +6,24 @@ nextflow.enable.configProcessNamesValidation = false
 
 docker.enabled = true
 
+/*
+Include a config file containing the default values for workflows
+*/ 
+includeConfig './default_params.config'
+
 process {
     cpus = 1
     memory = 5.GB
     container = 'enzymefunctioninitiative/efi-est:latest'
     containerOptions = "-v $projectDir:$projectDir -v $projectDir/../shared:$projectDir/../shared -v $projectDir/../../lib:$projectDir/../../lib -v $EFI_DATA_DIR:$EFI_DATA_DIR"
+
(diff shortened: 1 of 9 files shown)


conf/docker.config
```diff
@@ -6,12 +6,27 @@ nextflow.enable.configProcessNamesValidation = false
...
+        memory = params.cdhit_memory_limit
+    }
+
+    withLabel: 'TASK_create_ssn' {
+        cpus = 1
+        memory = '20 GB'
+    }
+
     withName: blastreduce {
         memory = 20.GB
```
nilsoberg, line 31: Is there a difference between `20 GB` and `20.GB`?

conf/docker.config
```diff
@@ -6,12 +6,27 @@ nextflow.enable.configProcessNamesValidation = false
...
+*/
+includeConfig './default_params.config'
+
 process {
     cpus = 1
     memory = 5.GB
     container = 'enzymefunctioninitiative/efi-est:latest'
     containerOptions = "-v $projectDir:$projectDir -v $projectDir/../shared:$projectDir/../shared -v $projectDir/../../lib:$projectDir/../../lib -v $EFI_DATA_DIR:$EFI_DATA_DIR"
 
+    withLabel: 'APP_cd_hit' {
```
nilsoberg, line 20: What's the difference between the `APP` and `TASK` prefix?

nilsoberg (approved):
Looks good.  I think in this case it would be good to run the entire test suite in deployment because of the change in the default params import in the config files, just to verify that nothing broke.

conf/docker.config
rbdavid: My understanding is, if we use single or double quotation marks (e.g. `'20 GB'`) around the "memory unit" value, it is parsed as a MemoryUnit object. Similarly, if there is a decimal point (e.g. `20.GB`) that is also parsed as a MemoryUnit object. So either work. I can follow the integer/non-string format to make it more standard with the rest of our configuration.

conf/docker.config
rbdavid: This is purely a naming convention I wanna use. 

The `APP_` prefix indicates a configuration label applied because the given app is called within the labeled process block. For example, I imagine a `'APP_duckdb'` label will also be created and applied to any process blocks that call duckdb, whether using the CLI or python interfaces. Similarly, other apps like BLAST, seqkit, Clustal, HMMER, Muscle, usearch, ShortBred (eventually), etc. Not all of these will need their own configuration block or special resource allocation, but they can all be labeled to denote what software tool is being used.

Similarly, the `TASK_` prefix would be used to label process blocks performing some specific task (e.g. "create_ssn") that require a special resource allocation. Since process blocks reading and writing ssns have been seen to require more memory, more wallclock time, and might need to be done in a perfectly parallel way, the `'TASK_create_ssn'` label should be applied to any process block that does those types of tasks. Other task labels can be added as needed.

rbdavid: In conversation, Nils has agreed this an appropriate structure to use.

PR #378, opened 2026-06-24
376 ssn performance optimization

### LINKED ISSUE #376: SSN performance optimization
Performance suffers when writing SSNs in the `generatessn`, `colorssn`, `gnt`, `clusteranalysis`, and `neighborhoodconnectivity` pipelines.  There are a few issues:

1. `generatessn` (using the `EFI::SSN::XgmmlWriter` module) is not performant because millions of edges are being created as objects before being written to disk.  Additionally, network latency creates a significant bottleneck for the millions of small write operations.
2. `colorssn`, `clusteranalysis`, `neighborhoodconnectivity`, and `gnt` are not performant because millions of edges are being constructed in memory.  The `XML::LibXML::Reader` module reads an edge, creates an object, then the module used by the four pipelines (`EFI::SSN::AttributeWriter`) constructs a data structure, passes that to `XML::Writer`, which creates XML elements in memory before writing.  This reconstruction and serialization is unnecessary and results in millions of write operations over a high-latency NFS filesystem.
3. `gnt` is additionally not performant because it inefficiently reads a GND sqlite file.  It uses uncached statement handles and queries the database for data from each individual sequence, resulting many millions of small read operations.

In a quick test, using node-local storage is not significantly more performant.  However, more testing in this area would be useful.

Optimize the pipelines as follows, in the following order with first being highest priority:

- [x] Use a bulk retrieval operation to obtain node attributes from the GND in one or two operations rather than in millions.  A quick test brought the run time in one case from a multi-day read operation to a matter of seconds.
- [x] In `EFI::SSN::AttributeWriter` (which uses `EFI::Xgmml::Writer`, a wrapper around `XML::Writer`), use the pass-through capability of `XML::Writer` to pass a raw edge string from the input file to the output file, without parsing and object construction.
- [x] In both modules that write XGMML files, implement an application-level memory buffer.  Instead of making `XML::Writer` responsible for streaming directly to the output file, make the code write to an in-memory scalar buffer, and manually flush the buffer in large sequential chunks that are optimized for NFS/ZFS.

These changes eliminate millions of network round-trips, allowing the ZFS storage system to handle data read/writes optimally.

**Proof of Concept**

```
my $memBuffer = "";
open my $memFh, ">", \$memBuffer;
open my $outputFh, ">", $outputFile;
$outputFh->autoflush(0);

my $writer = new XML::Writer(...OUTPUT => $memFh);

while (process_code()) {
    # Do xml stuff
    if (length($memBuffer) >= MEM_BUF_SIZE) {
        $outputFh->print($memBuffer);
        $memBuffer = "";
        open $memFH, ">", \$memBuffer;
    }
}

if (length($memBuffer) > 0) {
    $outputFh->print($memBuffer);
}

# Cleanup and close handles
````

### PR DESCRIPTION
**Pull request by nilsoberg**
The core of this commit adds optimizations to XGMML generation:

* Use string buffer to buffer writes to output files.  This reduces the amount of round-trip writes done over a network file system potentially from tens of millions into a much smaller set of 10MB chunks
* Use passthrough when writing edges in the AttributeWriter modules (Color SSN, Cluster Analysis, Neighborhood Connectivity GNT); these don't need to be parsed and constructed into objects when reading from the file since they are immediately written to the output file without updating the object
* Bulk retrieve from GND in the GNT color SSN process.  This eliminates millions of small reads which can overwhelm a network file system

Additionally:

* A bug was fixed in the EFI::Sequence module when getting an attribute value
* The annotation/metadata file separator character, used to separate multiple records for a single node attribute, was changed from the ASCII record separator character `1f` to `^`.  The character was causing a problem
* Node attributes passed to XGMML are filtered to remove non-printable characters

### CODE REVIEW

File: lib/EFI/GNT/GND/Reader.pm [hunk 1]
```diff
@@ -113,7 +116,10 @@ sub getAttribute {
     my ($colName, $sql) = $self->getAttributeSql($attr);
     return "" if not $colName;
 
-    my $sth = $self->{dbh}->prepare($sql);
+    my $sth = $self->{_sth}->{$sql};
```
nilsoberg, line 119: Cache the prepared statement handle in case it gets reused.

lib/EFI/GNT/GND/Reader.pm
```diff
@@ -124,6 +130,34 @@ sub getAttribute {
 }
 
 
+# public
+sub getAllGntNeighborData {
+    my $self = shift;
+
+    my $nbTable = NEIGHBOR_TABLE;
+    my $queryTable = QUERY_TABLE;
+    my $sql = "
```
nilsoberg, line 139: This bulk retrieval obtains all of the nodes, number of neighbors, ENA ID, and the families for all neighbors in one query.  This is much faster than individual queries for each node.

pipelines/generatessn/annotations/get_annotations.pl
```diff
@@ -197,7 +197,7 @@ sub formatAnnoData {
         # If the field doesn't exist in the database attributes but exists in the existing metadata
         # then use the existing value
         } elsif (not $data->{$field}) {
-            my $value = $inputIds->getSequence($accession)->getAttribute($field);
+            my $value = $inputIds->getSequence($accession)->getAttribute($field, 1);
```
nilsoberg, line 200: Expand multiple records for one node into an array ref

pipelines/generatessn/annotations/get_annotations.pl
```diff
@@ -186,7 +186,7 @@ sub formatAnnoData {
         # Set the UniRef cluster sequence IDs
         if ($field eq $clusterField) {
             my @ids = map { $_->[0] } @{$unirefIds->{$accession}};
-            $data->{$field} = join(",", $accession, @ids);
+            $data->{$field} = join(ANNO_ROW_SEP, $accession, @ids);
```
nilsoberg, line 189: Ensure all metadata is now separated by the global record separator

lib/EFI/Sequence.pm
```diff
@@ -48,8 +48,7 @@ sub getAttribute {
     my $val = $self->{attr}->{$attr};
     return "" if not $val;
     if ($doUnpack) {
-        my @vals = $self->unpackAttributeValue($val);
-        return \@vals;
+        return $self->unpackAttributeValue($val);
```
nilsoberg, line 51: `unpack` already takes into account the `wantarray` logic

lib/EFI/SSN/XgmmlWriter.pm
```diff
@@ -346,15 +346,9 @@ sub makeNodeAttributes {
...
-        # If the value is a scalar, but the field type is a list, then split the value into pieces
-        # to force the values into a XGMML list.  This is done because database fields with
-        # multiple values are separated by commas.
-        if ($fields->{$field}->{is_list}) {
-            my @vals = ref $value ? @$value : ($value);
-            $value = [ map { split(m/[,]/) } @vals ];
         # If the value is an array, and is only one, then save it as a scalar because we only want
         # to force XGMML list type if it is a true array
-        } elsif (ref $value eq "ARRAY" and @$value == 1) {
+        if (not $fields->{$field}->{is_list} and (ref $value eq "ARRAY" and @$value == 1)) {
```
nilsoberg, line 359: If the node attribute type is not a list type, and the value is an array ref, and has only one element, we don't want to convert to a list attribute (which will happen if an array ref is passed to the XML code).

lib/EFI/SSN/AttributeWriter.pm
```diff
@@ -230,22 +230,15 @@ sub copyElementWithoutNamespace {
...
-            push @attr, $attrName => $attrValue if $attrValue;
-        }
 
-        if ($self->{reader}->isEmptyElement()) {
-            $self->emptyTag("edge", @attr);
-        } else {
-            $self->startTag("edge", @attr);
-        }
-    } elsif ($self->{reader}->nodeType == XML_READER_TYPE_END_ELEMENT) {
-        $self->endTag("edge");
```
rbdavid, line 247: We don't need this elsif block since we are just copying the file from input into output?

lib/EFI/SSN/XgmmlWriter.pm
rbdavid: But don't we still need to handle the instance where a field is a list with more than one element, where we want the value to be `[value1,value2]`?

lib/EFI/Xgmml/Writer.pm
```diff
@@ -19,8 +19,11 @@ sub new {
     my $self = {};
     bless($self, $class);
 
+    $self->{xmlns} = $args{xmlns} // DEFAULT_XMLNS;
     $self->{data_indent} = $args{data_indent} // 0;
     $self->{output_file} = $args{output_file};
+    $self->{mem_buffer_size} = 10 * 1024 * 1024;
```
rbdavid, line 25: Should this be saved as a constant elsewhere or as an input argument in case we want to handle the buffer size dynamically? Just a question.

pipelines/gnt/color_gnt_xgmml.pl
```diff
@@ -112,30 +112,13 @@ sub getGntData {
     my $gnd = new EFI::GNT::GND::Reader();
     $gnd->load($gndFile);
 
-    my $gntData = {};
-    foreach my $cluster ($gnd->getClusterNums()) {
-        my @queryIds = $gnd->getQueryIds($cluster);
-        foreach my $queryId (@queryIds) {
-            my $data = {};
-            $data->{ena_id} = $gnd->getAttribute($queryId, ATTR_QUERY|ATTR_ENA_ID);
```
rbdavid, line 120: Are we not gathering the `ena_id` attribute anymore? Was this data ever incorporated into `$gntData` array?

rbdavid (approved):
What tests should I run to verify these changes work as expected? And which workflows will see performance improvements?

lib/EFI/SSN/XgmmlWriter.pm
nilsoberg: If the field is a list with more than one element, then it is made into a list in the code that writes the node attributes.  The `@$value == 1` check ensures that this is only run for arrays that are exactly size 1 and are not a list type.

lib/EFI/SSN/AttributeWriter.pm
nilsoberg: Correct

lib/EFI/Xgmml/Writer.pm
nilsoberg: Could be.

nilsoberg: I'm electing to leave it as a constant in this file.  If it becomes an issue in the future we can add a parameter.  As it stands 6 pipelines and 4 scripts would have to be adjusted to take this as a new parameter.  Would like to minimize changes at this point.

pipelines/gnt/color_gnt_xgmml.pl
nilsoberg: That comes in the SQL in `EFI::GNT::GND::Reader` on line 142.

lib/EFI/SSN/XgmmlWriter.pm
```diff
@@ -160,14 +161,12 @@ sub writeNode {
 
             my @values;
             if (ref $value eq "ARRAY") {
-                @values = map { ref $_ eq "ARRAY" ? @$_ : $_ } @$value;
+                @values = @$value;
```
nilsoberg, line 164: A previous change ensured there were no nested array refs.

lib/EFI/Annotations.pm
```diff
@@ -272,42 +321,6 @@ sub parse_interpro {
 }
 
 
-#
```
nilsoberg, line 273: Moved up to put it with related functions.

rbdavid (approved):

PR #379, opened 2026-06-24
377 - axa blast configuration

### LINKED ISSUE #377: All-by-All Blast configuration
- [x] Timescaling of blast on multiple cpus. Explicitly define nProcessors (via `-a` input arg). 
- [x] Alternatively, make the num_fasta_shards number dynamically change to maintain a reasonable amount of sequences per shard to avoid memory limitations. 
- [x] Copy the outputs of formatdb call into a compute node-local storage space. Write output file to that same storage space and then copy to publishDir when completed.

rbdavid:
Closes #377 

Implements configuration control of process blocks that call blastp.


nilsoberg (approved):
Looks fine

nilsoberg:
Are you using `APP_blastp` for both the process of collection sequences (Option A) and the all-by-all?  It might be beneficial to have two separate blocks.  The initial collection could potentially benefit from using threads, while the all-by-all shouldn't use threads (at least in the way it is currently configured).

rbdavid:
The `APP_blastp` call uses the default `params.blastp_threads` (newly added) to control the cpu allocation for each labeled process block, and that default value can be changed depending on where/how the backend codes are implemented. 

In my next PR, the `all_by_all_blast` process block will also be labeled with `APP_duckdb`. The `APP_duckdb` configuration, being processed after the `APP_blastp` configuration, will take priority and allow us to adequately allocate for both the blastp and duckdb resources. For each implementation (slurm, pbspro, etc), the label configurations can be changed to best fit the deployment. That being said, I believe duckdb has higher memory overhead and more efficiently utilizes more CPUs, so I think its smart to prioritize the duckdb configuration over the blastp configuration (this is bleeding into the duckdb config PR that I've got waiting in the wings). 

I don't really understand why the number of CPUs will affect the `all_by_all_blast` process or why that process block shouldn't be multi-threaded. Once again, if "differences" in blast results occur due to ordering of sequences or memory handling or multi-threading effects, these will be systematic differences (issues arising due to the blastall software). Differences will only be seen when comparing between two differently allocated calls of the `blastall -p blastp` command.

nilsoberg (changes requested):
It's great that these problem areas have been identified.  I am a bit concerned that we're over-complicating the config files.  Rather than lock ourselves into larger resource allocations, individual processes (such as the one you identified in the blast input path) need to be profiled and optimized at a low level.  I'll think about it some more next week, so let's hold off on merging this PR.  It would be good to see any data that you have collected.

rbdavid:
Table for BLAST import mode, all sequence database versions/sources shown with input sequence being constant:

| Sequence Database | Process Name | Elapsed Time | CPU Usage | Peak RSS | Peak Virtual Mem | Read | Write |
|-------------------|--------------|--------------|-----------|----------|-----------------|------|-------|
| UniRef50 | IMPORT_AND_FILTER:GET_SOURCE_IDS:get_source_ids_blast (1) | 7m 23s | 98.9% | 15.3 GB | 17.8 GB | 1.2 MB | 2.3 KB |
| UniRef90 | IMPORT_AND_FILTER:GET_SOURCE_IDS:get_source_ids_blast (1) | 20m 12s | 99.4% | 45.1 GB | 52.5 GB | 1.2 MB | 2.4 KB |
| UniProt | IMPORT_AND_FILTER:GET_SOURCE_IDS:get_source_ids_blast (1) |  44m 34s | 83.2% | 84.8 GB | 129.9 GB | 1.2 MB | 3.2 KB |

rbdavid:
For all my Blast import mode tests, the `get_source_ids_blast` process block is above and beyond the computational bottleneck.

nilsoberg (approved):

PR #381, opened 2026-06-25
376 ssn performance optimization buffer

### LINKED ISSUE #376: SSN performance optimization

nilsoberg:
This PR optimizes parts of the generatessn, clusteranalysis, neighborhoodconnectivity, colorssn, and gnt pipelines by the following:

* Undoing a change designed to mitigate issues arising from NFS (removing the string buffer code added in a previous branch)
* Avoiding storing edgelists in memory and instead writing directly to disk
* Flattens metadata structures when parsing SSNs

### DIFF
--- a/lib/EFI/SSN/XgmmlReader.pm
+++ b/lib/EFI/SSN/XgmmlReader.pm
@@ -14,7 +14,6 @@ sub new {
     bless($self, $class);
 
     $self->{domain_idx} = {};
-    $self->{edgelist} = [];
     $self->{input} = $args{xgmml_file};
     $self->{id_idx} = {};
     $self->{idx_seqid} = {};
@@ -24,12 +23,6 @@ sub new {
     $self->{metadata_only} = $args{metadata_only} // 0;
 
     return $self;
-}
-
-
-sub getEdgeList {
-    my $self = shift;
-    return $self->{edgelist};
 }
 
 
@@ -59,56 +52,44 @@ sub getDomainIndexMap {
 
 sub parse {
     my $self = shift;
+    my %args = @_;
 
     my $reader = XML::LibXML::Reader->new(location => $self->{input}) or die "cannot read $self->{input}\n";
     $self->{current_node_id} = "";
-    while ($reader->read) {
-        my $continueParsing = $self->processXmlNode($reader);
-        last if not $continueParsing;
+
+    my $id_idx = $self->{id_idx};
+
+    my $edgelistFh;
+    if ($args{edgelist_file}) {
+        open $edgelistFh, ">", $args{edgelist_file} or die "Unable to open edgelist file '$args{edgelist_file}' for writing: $!";
     }
-}
 
+    my $TYPE_ELEMENT = XML_READER_TYPE_ELEMENT;
 
-#
-# processXmlNode - private method
-#
-# Processes a XML node (a XGMML 'edge', 'node', or 'att' tag). Called for every
-# type of XML element encountered, but only the start node or empty nodes are
-# processed.
-#
-# Parameters:
-#    $reader - XML::LibXML::Reader object (points to current XML node)
-#
-# Returns:
-#    non-zero if ok to continue parsing, zero to terminate parsing
-#
-sub processXmlNode {
-    my $self = shift;
-    my $reader = shift;
-    my $ntype = $reader->nodeType;
-    my $nname = $reader->name;
+    while ($reader->read) {
+        my $ntype = $reader->nodeType;
+        next if $ntype != $TYPE_ELEMENT;
 
-    return 1 if $ntype == XML_READER_TYPE_WHITESPACE || $ntype == XML_READER_TYPE_SIGNIFICANT_WHITESPACE;
+        my $nname = $reader->name;
 
-    if ($ntype == XML_READER_TYPE_ELEMENT) {
-        if ($nname eq "node") {
+        if ($nname eq "edge") {
+            my $source = $reader->getAttribute("source");
+            my $target = $reader->getAttribute("target");
+            $edgelistFh->print($id_idx->{$source}, " ", $id_idx->{$target}, "\n");
+        } elsif ($nname eq "node") {
             $self->processNode($reader);
         } elsif ($nname eq "att") {
             # An 'empty' element is a leaf (e.g. no child elements; <att X="Y" /> is empty)
             if ($reader->isEmptyElement()) {
                 $self->processAtt($reader);
             }
-        } elsif ($nname eq "edge") {
-            $self->processEdge($reader);
         } elsif ($nname eq "graph") {
             $self->processGraph($reader);
-            if ($self->{metadata_only}) {
-                return 0;
-            }
+            last if $self->{metadata_only};
         }
     }
 
-    return 1;
+    close $edgelistFh if $edgelistFh;
 }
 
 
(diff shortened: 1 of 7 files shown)


lib/EFI/SSN/XgmmlReader/IdList.pm
```diff
@@ -24,7 +24,7 @@ sub new {
     $self->{anno} = new EFI::Annotations;
     my ($attrNames, $attrDisplay) = $self->{anno}->get_expandable_attr();
     $self->{id_list_fields} = { map { $attrDisplay->{$_} => $_ } @$attrNames };
-    $self->{id_metadata} = {}; # any node metadata that we are to store (e.g. swissprot)
+    $self->{sequences} = {}; # any custom sequences to store
```
nilsoberg, line 27: Removed swissprot and only storing sequences (e.g. from Option C).  Swissprot annotations are retrieved elsewhere.

lib/EFI/SSN/XgmmlReader/IdList.pm
```diff
@@ -134,12 +117,10 @@ sub processNodeAttribute {
 
         # Store the value in a hash ref in the case that the network is UniRef+RepNode
         # (in that case there will be duplicates because of the FIELD_REPNODE_IDS values)
-        $self->{meta_map}->{$seqId}->{$value} = 1;
-    # SwissProt
```
nilsoberg, line 138: Removed swissprot and only storing sequences (e.g. from Option C). Swissprot annotations are retrieved elsewhere in other processes.

lib/EFI/SSN/XgmmlReader.pm
```diff
@@ -59,56 +52,44 @@ sub getDomainIndexMap {
 
 sub parse {
```
nilsoberg, line 53: There were two primary optimizations in this file:

1. Writing edges directly to the edgelist file instead of storing in an array ref then saving in the calling script.  This reduced the amount of RAM used by several orders of magnitudes in large tests. It also resulted in a mild improvement in runtime.
2. Since this loop is executed potentially hundreds of millions of times (num_nodes X num_attributes + num_edges * 4, at a minimum), avoiding unnecessary function calls by inserting the contents of smaller functions directly in the `parse` function reduces overhead.

lib/EFI/Xgmml/Writer.pm
```diff
@@ -40,57 +38,18 @@ sub open {
         flock($fh, LOCK_EX) or warn "Unable to obtain exclusive file lock on output SSN for writing: $!";
     };
 
-    # Pass $self into the writer, which will call print on the handle (see below)
-    my $writer = new XML::Writer(OUTPUT => $self, DATA_INDENT => $self->{data_indent}, UNSAFE => 1, PREFIX_MAP => '');
+    my $writer = new XML::Writer(OUTPUT => $fh, DATA_INDENT => $self->{data_indent}, UNSAFE => 1, PREFIX_MAP => '');
```
nilsoberg, line 41: This change removes the string buffer optimization.

pipelines/gnt/color_gnt_xgmml.pl
nilsoberg, line 1: Time reporting is retained in case there is further room for optimization later.

pipelines/shared/perl/compute_conv_ratio.pl
```diff
@@ -25,14 +25,14 @@
 # Get the metanode data (mapping of repnode/UniRef to UniProt)
 my ($idType, $sourceIdMap) = parse_metanode_map_file($opts->{seqid_source_map});
 
-# Get the edgelist
-my $edgelist = parseEdgelist($opts->{edgelist});
+# Get the node degrees from the edgelist
```
nilsoberg, line 28: The edgelist is no longer returned.  `computeDegrees` was computing the node degree based on the edge list.  Rather than store an edgelist in memory, node degrees are computed when the edgelist file is loaded.  In one test this reduced the memory usage a factor of 90.

pipelines/shared/perl/ssn_to_id_list.pl
```diff
@@ -41,8 +38,8 @@
 saveMetanodeMapping($opts->{seqid_source_map}, $metanodeMap, $metanodeType);
 
 if ($opts->{ssn_sequences}) {
-    my $metadata = $parser->getMetadata();
-    saveSsnSequences($opts->{ssn_sequences}, $metadata);
+    my $sequences = $parser->getSequences();
```
nilsoberg, line 41: Using the purpose-built data structure for SSN sequences

lib/EFI/SSN/XgmmlReader/IdList.pm
```diff
@@ -134,12 +117,10 @@ sub processNodeAttribute {
...
-    } elsif ($fieldName eq FIELD_SWISSPROT_DESC) {
-        $self->{id_metadata}->{$seqId}->{swissprot} = $value if $value;
+        $self->{meta_map}->{$seqId}->{$value} = undef;
```
rbdavid, line 120: Why `undef` instead of a `1`? Memory consideration or some other reason?

lib/EFI/SSN/XgmmlReader/IdList.pm
```diff
@@ -71,26 +71,9 @@ sub getMetanodes {
 }
 
 
-sub getMetadata {
```
rbdavid, line 74: Was `id_metadata` only ever used to stash the sequence?

pipelines/shared/perl/compute_conv_ratio.pl
rbdavid: I was using duckdb to calculate degree centrality for nodes when I was looking for the source of extra edges a few weeks ago. It was very quick for calculating those values. Maybe we should use duckdb for edge analyses? Just an idea.

rbdavid: Maybe that's not possible when metanodes and other complications are considered.

rbdavid (approved):
Looks good. Any chance we can, minimize data structures to nNodes instead of nEdges (when possible) is the best way to decrease memory overhead. 

But still, using memory should be fine and much faster than the amount of IO calls that are necessary to avoid the memory overhead by writing files to storage. Along with the common sense changes here, I think a bit more profiling of the workflows and making the processes' resource allocations more dynamic would enable us to utilize memory efficiently even on systems where memory may be limited.

lib/EFI/SSN/XgmmlReader/IdList.pm
nilsoberg: Memory usage

lib/EFI/SSN/XgmmlReader/IdList.pm
nilsoberg: Yes

PR #386, opened 2026-06-29
340 descriptively name ssns in ssn utility output

### LINKED ISSUE #340: Descriptively name SSNs in SSN Utility output
SSNs output from NC utility (and maybe others) are named `ssn.xgmml`.  Name these based on the input file name or job name.

nilsoberg:
This PR contains the following changes:

* SSNs output from generatessn, nf, clusteranalysis, colorssn, and gnt are now named appropriately
* Zipping of the SSN is accomplished within the same process block as creation (so that the output name can be consistent in the process output block)
* This also closes the file capitalization issue that has been tracked
* Additionally, the code for coloring SSNs as part of the generatessn pipeline has been removed; this was not used


pipelines/generatessn/generatessn.nf
```diff
@@ -359,10 +348,4 @@ workflow {
 
     // Merge full and repnode SSN stats into one file
     final_stats = merge_stats(full_ssn.stats.mix(final_ids.import_stats, repnode_stats).collect())
-
-    if (params.color_ssn) {
-        computed = COMPUTE_COLOR_CLUSTER_WORKFLOW(full_ssn.ssn_unzipped)
-        colored_ssn = color_ssn(full_ssn.ssn_unzipped, computed.cluster_id_map, computed.cluster_num_map, computed.cluster_colors)
-        zipped_full_ssn = zip_files(colored_ssn.ssn)
```
rbdavid, line 366: Is this an optimization change? Instead of just creating the SSN and repnodes, the current workflow was also creating a colored SSN, so removing this step will inherently speed up the `generatessn.nf` workflow. But, I thought this was a good idea to implement since colored SSNs are more visualization-ready than the base SSNs output from the `generatessn.nf` code.

rbdavid (approved):
I guess I thought having `generatessn.nf` produce colored SSNs was a good idea, as it would save an analysis step before submitting GNTs and output SSNs that were more readily visualized. Why are we removing that call? 

Everything else makes sense and looks good to go, after testing.

pipelines/generatessn/generatessn.nf
nilsoberg: This feature is not currently used on the production site.  The implementation as coded conflicted with the neighborhood connectivity option.  Additional validation would be necessary to ensure that the the NC feature didn't conflict with the color SSN option.  Additionally, one of the important features of the color SSN pipeline that is valuable is the ancillary files (e.g. ID lists).  The color SSN would have to be recolored to get this data.  Finally, in order to color a SSN in the generatessn pipeline an additional process would have to have been added to do the unzipping before coloring (or an additional process would have to be added to do file renaming before zipping, but that would involve making a copy of the file, renaming it, then zipping it up; this is inefficient for large files).  Ultimately it comes down to stakeholders.  It is easy to add this back into the code should it be needed in the future.

PR #388, opened 2026-06-30
387 copy seq len node attributes for zz sequences

### LINKED ISSUE #387: Copy seq len node attributes for ZZ sequences
When preparing annotations in `get_annotation.pl` for SSN creation, ensure that the source `seq_len` attribute is copied from the source metadata file.

nilsoberg:
This PR contains changes that address issues with SSNs:

1. Sequence lengths are now included for FASTA jobs
2. BLAST job input sequence length is now included
3. Validation now occurs to ensure that non-integer values never get saved into an integer field (Cytoscape crashes when that occurs)

### DIFF
--- a/lib/EFI/Import/Source/FASTA.pm
+++ b/lib/EFI/Import/Source/FASTA.pm
@@ -11,10 +11,9 @@ use Data::Dumper;
 use Cwd qw(abs_path);
 use File::Basename qw(dirname);
 use lib dirname(abs_path(__FILE__)) . "/../../../"; # Import libs
-use lib dirname(abs_path(__FILE__)) . "/../../../../../../../lib"; # Global libs
 use parent qw(EFI::Import::Source);
 
-use EFI::Annotations::Fields ':source';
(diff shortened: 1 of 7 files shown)


lib/EFI/Import/Source/FASTA.pm
```diff
@@ -229,7 +228,10 @@ sub makeMetadata {
     my $destSeqData = shift;
 
     foreach my $id (keys %$seq) {
-        my $attr = { &FIELD_SEQ_SRC_KEY => FIELD_SEQ_SRC_VALUE_FASTA };
+        my $attr = {
+            &FIELD_SEQ_SRC_KEY => FIELD_SEQ_SRC_VALUE_FASTA,
```
rbdavid, line 232: Should `FIELD_SEQ_SRC_KEY` be called in line 16?

rbdavid (approved):
Looks good.

lib/EFI/Import/Source/FASTA.pm
nilsoberg: Line 16 imports the constant.

PR #397, opened 2026-07-13
395 - Muscle process block configuration

### LINKED ISSUE #395: Muscle configuration
Muscle calls need to be parallelized and resource allocations need to be set.

rbdavid:
- [x] Fix errorStrategy for muscle3_process block.
- [x] Configure the muscle process blocks to hard-set the maxtime to limit how long these process types can run. 
- [x] Unfortunately, no control for RAM usage is possible but still implemented a maximum for allocated RAM too.

Closes #395

### DIFF
--- a/conf/docker.config
+++ b/conf/docker.config
@@ -55,6 +55,7 @@ process {
 
     withLabel: muscle_align {
         cpus = 1
+        time = 24.hr
     }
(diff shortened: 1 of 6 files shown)


nilsoberg (approved):
Looks good

PR #398, opened 2026-07-13
396 - Add label to the GNTGNN generation process block

### LINKED ISSUE #396: SSN creation task configuration
Nextflow configuration of process blocks that create SSN files needs to be developed.

rbdavid:
Handle GNN creation process block similar as the SSN creation process block. Control for large required memory allocation and potentially large IO amount.


rbdavid:
Closes #396

nilsoberg (approved):

PR #399, opened 2026-07-13
349 compute cr for uniref in cr utility

### LINKED ISSUE #349: Compute CR for UniRef in CR utility
When using a UniRef-based SSN as the input to the convergence ratio utility, compute convergence ratio for both UniProt and UniRef.

nilsoberg:
This PR fixes issues in the convergence ratio pipeline.  Previously, the "SSN convergence ratio" value that was reported was the same value computed by the color SSN, which always based on the UniRef nodes (e.g. UniRef clusters are expanded).  However, in the legacy code for the convergence ratio utility, this ratio is computed based on the number of metanode edges for UniRef inputs.  A flag was added that conditionally enabled computation based on metanode rather than UniProt sequences, for the convergence ratio pipeline.  The color SSN pipeline retains the same functionality.  Additionally, due to the UniProt database changes, some clusters may end up having no sequences identified when retrieving FASTA.  The updates to the code ensure that these clusters are excluded from the computation.  Finally, in the python code for computing convergence ratio, proper sorting was implemented as was support for cluster IDs (e.g. `Cluster_1` rather than the previous numeric value of `1`).

### DIFF
--- a/pipelines/convergenceratio/convergenceratio.nf
+++ b/pipelines/convergenceratio/convergenceratio.nf
@@ -1,7 +1,7 @@
 
 include { all_by_all_blast; blastreduce; blastreduce_transcode_fasta; condense_redundant; create_blast_db; restore_condensed; sort_and_split_fasta } from "../shared/nextflow/blast.nf"
 include { unzip_ssn } from "../shared/nextflow/util.nf"
-include { compute_clusters; get_conv_ratio_table; get_id_list; get_ssn_id_info } from "../shared/nextflow/color_workflow.nf"
+include { compute_clusters; get_id_list; get_ssn_id_info } from "../shared/nextflow/color_workflow.nf"
 
 process compute_blast_conv_ratio {
     input:
@@ -56,25 +56,45 @@ process get_fasta_files {
     """
 }
 
-process get_selected_id_lists {
+process compute_ssn_conv_ratio {
     input:
-        path uniprot_dir
-        path uniref90_dir
-        path uniref50_dir
-
+        path edgelist
+        path index_seqid_map
+        path cluster_id_map
+        path seqid_source_map
     output:
-        path "selected_ids/*.txt", emit: "id_files"
+        path "conv_ratio.txt", emit: conv_ratio
 
     script:
     """
-    mkdir -p selected_ids
+    perl $projectDir/../shared/perl/compute_conv_ratio.pl \
+        --cluster-map $cluster_id_map \
+        --index-seqid-map $index_seqid_map \
+        --edgelist $edgelist \
+        --seqid-source-map $seqid_source_map \
+        --conv-ratio conv_ratio.txt \
+        --use-metanodes
+    """
+}
 
-    if ls ${uniref50_dir}/*.txt 1> /dev/null 2>&1; then
-        cp ${uniref50_dir}/*.txt selected_ids/
-    elif ls ${uniref90_dir}/*.txt 1> /dev/null 2>&1; then
-        cp ${uniref90_dir}/*.txt selected_ids/
+process count_fasta {
+    tag "ca_count_fasta_${id}"
+
+    input:
+        tuple val(id), path(fasta)
+
+    output:
+        // We emit the same tuple structure so it can mix back easily later, and add the number
+        // of sequences in the input file to the tuple
+        tuple val(id), path(fasta), env(COUNT)
+
+    script:
+
+    """
+    if [ -s "${fasta}" ]; then
+        COUNT=\$(grep -c "^>" ${fasta})
     else
-        cp ${uniprot_dir}/*.txt selected_ids/
+        COUNT=0
     fi
     """
 }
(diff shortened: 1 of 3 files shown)


pipelines/shared/perl/compute_conv_ratio.pl
```diff
@@ -106,13 +107,14 @@ sub computeConvRatio {
...
+        my @clusterIds = $useMetanodes ? @{ $clusterToId->{$cnum} } : @{ $fullClusterToId->{$cnum} };
         my $numDegree = 0;
-        foreach my $id (@fullIds) {
+        foreach my $id (@clusterIds) {
             next if not $degrees->{$id};
             $numDegree += $degrees->{$id};
         }
 
-        my $denom = $numIds * ($numIds - 1);
+        my $denom = $useMetanodes ? $numNodes * ($numNodes - 1) : $numIds * ($numIds - 1);
```
rbdavid, line 118: I think it is important to note that you are double counting the edges between nodes on both the numerator and denominator, so those multiplications of 2 cancel out.

pipelines/convergenceratio/convergenceratio.nf
```diff
@@ -56,25 +56,45 @@ process get_fasta_files {
...
     script:
     """
-    mkdir -p selected_ids
+    perl $projectDir/../shared/perl/compute_conv_ratio.pl \
+        --cluster-map $cluster_id_map \
+        --index-seqid-map $index_seqid_map \
+        --edgelist $edgelist \
+        --seqid-source-map $seqid_source_map \
+        --conv-ratio conv_ratio.txt \
+        --use-metanodes
```
rbdavid, line 73: `--use-metanodes` is always flagged as True?

pipelines/convergenceratio/convergenceratio.nf
```diff
@@ -56,25 +56,45 @@ process get_fasta_files {
...
-        cp ${uniref90_dir}/*.txt selected_ids/
+    """
+    if [ -s "${fasta}" ]; then
+        COUNT=\$(grep -c "^>" ${fasta})
     else
-        cp ${uniprot_dir}/*.txt selected_ids/
+        COUNT=0
     fi
     """
 }
```
rbdavid, line 93: I foresee this process block being actually quite useful. Could it be moved into a shared nextflow file? For example, if we wanted to dynamically scale the number of fasta shards to be created before running the axa blast process block, this count of sequences in the fasta file could be useful.

pipelines/convergenceratio/convergenceratio.nf
```diff
@@ -56,25 +56,45 @@ process get_fasta_files {
...
+        --index-seqid-map $index_seqid_map \
+        --edgelist $edgelist \
+        --seqid-source-map $seqid_source_map \
+        --conv-ratio conv_ratio.txt \
+        --use-metanodes
+    """
+}
+
+process count_fasta {
+    tag "ca_count_fasta_${id}"
```
rbdavid: What's this tag being used for?

pipelines/convergenceratio/convergenceratio.nf
rbdavid, line 162: It still doesn't make a whole bunch of sense to me why we re-run the axa blast processes here. The edge values are stashed in the SSN xgmml file that is uploaded _with the original alignment score already applied on those edges_. We could just parse the xgmml file for edge information rather than re-running blast.

pipelines/shared/perl/compute_conv_ratio.pl
```diff
@@ -34,10 +34,9 @@
 # Compute degrees, metanode only
 my $degrees = computeDegrees($nodeDegrees, $indexMap);
 
-# Expand the metanodes
-my $fullClusterToId = resolve_mapping($clusterToId, $idType, $sourceIdMap);
+my $fullNetworkIds = resolve_mapping($clusterToId, $idType, $sourceIdMap);
```
rbdavid, line 38: Why was the comment removed?

pipelines/shared/perl/compute_conv_ratio.pl
```diff
@@ -276,6 +279,12 @@ =head3 Arguments
 to sequence IDs that are within the metanode. If present the convergence ratio is calculated
 by taking into account the full cluster size.
 
+=item C<--use-metanodes>
+
+If the input is an UniRef SSN and this flag is provided, then metanodes (e.g. SSN nodes) are used
+to compute the convergence ratio.  Otherwise the UniProt sequences in the cluster are used.  If
+the input is UniProt, then the number of sequences are equivalent to the number of nodes.
```
rbdavid, line 287: The flag is always set at the moment. So, if a UniRef SSN is uploaded, the two convergence ratios will be equivalent because the first will be for the clusters as seen in the uploaded file (like normal) and the second will be the same value because this flag is set and the metanodes are the nodes seen in the SSN file? Am I confused?

rbdavid (approved):
I'm not sure we're doing Convergence Ratio calculations the best way. But this is recreating the production site's method so that's fine by me for now.

pipelines/convergenceratio/convergenceratio.nf
nilsoberg: `--use-metadnodes` is not used when computing convergence ratio in the color ssn pipeline.

pipelines/convergenceratio/convergenceratio.nf
nilsoberg: That came from the cluster analysis pipeline and doesn't need to be tagged here.

pipelines/convergenceratio/convergenceratio.nf
nilsoberg: Yes, it could be reused; in fact, similar code is used in the cluster analysis pipeline (`subworkflows/prepare.nf`).  The difference there is the tuple input and output.  I would say that once we figure out how to use this in the fasta scaling code, we can refactor at that point and code it the way it makes most sense (e.g. with the tuple input/output).

pipelines/shared/perl/compute_conv_ratio.pl
nilsoberg: Accident

pipelines/shared/perl/compute_conv_ratio.pl
nilsoberg: It's not quite that simple.  The first CR calculated represents the convergence ratio that was computed after performing the all-by-all BLAST and is dependent on the input e-value.  The second CR calculated is theoretical and compares the actual SSN edges with the number of edges possible.  Therefore, both values will most likely be different, even if slightly.

PR #403, opened 2026-07-14
394 analysis filtering not properly applied

### LINKED ISSUE #394: Analysis filtering not properly applied
When applying filtering in the `generatessn` pipeline (i.e. Analysis step), UniRef cluster IDs are incorrect when length limits are filtered:

* For UniRef50 (analysis job 964), UniRef50 Cluster Size is 1 for all nodes for full SSN; Cluster IDs are just the Cluster ID.  For 85% rep node, it looks like the cluster size is always 1 but the Cluster IDs are correct (at least there is more than one value)
* The same for UniRef90.  For UniRef90 (analysis job 976), UniRef50 Cluster Size is 1 for all nodes for full SSN; Cluster IDs are just the Cluster ID.  For 85% rep node, it looks like the cluster size is always 1 but the Cluster IDs are correct.

This issue does not occur in analysis step filtering for Exclude Fragment and Taxonomy filtering.

Additionally, unlike the production site, the sizes of the UniRef clusters (and their internal IDs?) do not reflect loss of nodes/IDs from the filtering.  The filtering should remove internal cluster IDs that do not satisfy the filtering criteria.  Looks like it is ignored with Length, Exclude Fragment, and Taxonomy (?) filtering. 934/964,966,967. 933/976,977,978.

Excluding fragments in the Analysis step does not change the Sequence Status node attribute for the UniRef clusters. "fragment" is still indicated in the Sequence Status node attribute.  934/967.  933/977.

nilsoberg:
This PR contains fixes to filtering performed when creating SSNs (i.e. during the generatessn pipeline).  The code now supports automatic determination of sequence type to fix UniRef metadata synchronization in accession table and sequence metadata files - necessary for SSN filtering.  Length filtering now works properly and is applied in two steps: A) the `compute_fasta_length` process to come up with a list of IDs--both unknown ZZZ and UniProt--that are within the specified length filter range, and B) the normal `min-seq-length` and `max-seq-length` filters passed to `filter_ids.pl`.  Since the job manager passes these parameters as `--min-length` and `--max-length` rather than `--filter min-seq-length...`, a minor update converts these parameters into `--filter` arguments in the process block.  Previously, the taxonomy min/max length filter was not being applied due to the `--min-length` and `--max-length` argument usage, and this PR contains the fix that will work for taxonomy in addition to generatessn.  Finally, if the computed dataset is too small, the plot generation process was failing due to not having data.  This PR also contains a fix that creates empty graphs if there is no data (rather than failing).

When deploying these changes, the Python venv will need to be updated with pip in order for the plot changes to take effect.

### DIFF
--- a/lib/EFI/Import/Config/Filter.pm
+++ b/lib/EFI/Import/Config/Filter.pm
@@ -76,8 +76,8 @@ sub validateOptions {
     $opts->{fraction} = ($filter->{fraction} and $filter->{fraction} =~ m/^\d+$/) ? $filter->{fraction} : 1;
     $opts->{remove_fragments} = exists $filter->{fragments} ? 1 : 0;
     $opts->{family_filter} = $filter->{family};
-    $opts->{min_seq_length} = ($filter->{min_seq_length} and $filter->{min_seq_length} =~ m/^\d+$/) ? $filter->{min_seq_length} : DEFAULT_MIN_SEQ_LENGTH;
-    $opts->{max_seq_length} = ($filter->{max_seq_length} and $filter->{max_seq_length} =~ m/^\d+$/) ? $filter->{max_seq_length} : DEFAULT_MAX_SEQ_LENGTH;
+    $opts->{min_seq_length} = ($filter->{"min-seq-length"} and $filter->{"min-seq-length"} =~ m/^\d+$/) ? $filter->{"min-seq-length"} : DEFAULT_MIN_SEQ_LENGTH;
+    $opts->{max_seq_length} = ($filter->{"max-seq-length"} and $filter->{"max-seq-length"} =~ m/^\d+$/) ? $filter->{"max-seq-length"} : DEFAULT_MAX_SEQ_LENGTH;
     my @taxFilterErrors = $self->parseTaxonomyFilterOptions($filter, $opts);
     push @errors, @taxFilterErrors;
 
--- a/lib/EFI/Sequence/Collection.pm
+++ b/lib/EFI/Sequence/Collection.pm
@@ -303,8 +303,6 @@ sub updateUnirefMetadata {
     foreach my $unirefId (keys %uniprotIds) {
         my @ids = sort @{ $uniprotIds{$unirefId} };
         my $size = @ids;
-        print "Setting metadata for $unirefId:\n";
-        print "\t", join(",", @ids), " $size\n";
         $self->{seq}->{$unirefId}->setAttribute($attrName, \@ids);
         $self->{seq}->{$unirefId}->setAttribute($sizeAttrName, $size);
     }
--- a/lib/EFI/Sequence/Collection.pm
+++ b/lib/EFI/Sequence/Collection.pm
@@ -108,33 +108,44 @@ sub removeSequence {
     # Now remove from the accession ID table.  Primary sequence type refers to the sequence
     # type of the sequence version (e.g. UniProt, UniRef)
 
-    # If the primary sequence type is UniRef50 and this ID is a UniRef50 ID then delete all
-    # UniRef90 and UniProt IDs that are in that UniRef50 cluster
-    if ($self->{sequence_version} eq SEQ_UNIREF50 and $self->{uniref50}->{$sequenceId}) {
+    # If this ID is a UniRef50 ID then delete all UniRef90 and UniProt IDs that are in that
+    # UniRef50 cluster
+    if ($self->{uniref50}->{$sequenceId}) {
         foreach my $ur90 (keys %{ $self->{uniref50}->{$sequenceId} }) {
             if ($self->{uniref90}->{$ur90}) {
                 foreach my $up (keys %{ $self->{uniref90}->{$ur90} }) {
                     delete $self->{uniprot}->{$up};
+                    delete $self->{seq}->{$up} if $self->{seq}->{$up};
                 }
             } else {
                 delete $self->{uniprot}->{$ur90};
+                delete $self->{seq}->{$ur90} if $self->{seq}->{$ur90};
             }
             delete $self->{uniref90}->{$ur90};
         }
         delete $self->{uniref50}->{$sequenceId};
     }
-    
-    # If the primary sequence type is UniRef90 and this ID is a UniRef90 ID then delete all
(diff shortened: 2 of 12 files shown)


lib/EFI/Import/Config/Filter.pm
```diff
@@ -76,8 +76,8 @@ sub validateOptions {
     $opts->{fraction} = ($filter->{fraction} and $filter->{fraction} =~ m/^\d+$/) ? $filter->{fraction} : 1;
     $opts->{remove_fragments} = exists $filter->{fragments} ? 1 : 0;
     $opts->{family_filter} = $filter->{family};
-    $opts->{min_seq_length} = ($filter->{min_seq_length} and $filter->{min_seq_length} =~ m/^\d+$/) ? $filter->{min_seq_length} : DEFAULT_MIN_SEQ_LENGTH;
-    $opts->{max_seq_length} = ($filter->{max_seq_length} and $filter->{max_seq_length} =~ m/^\d+$/) ? $filter->{max_seq_length} : DEFAULT_MAX_SEQ_LENGTH;
+    $opts->{min_seq_length} = ($filter->{"min-seq-length"} and $filter->{"min-seq-length"} =~ m/^\d+$/) ? $filter->{"min-seq-length"} : DEFAULT_MIN_SEQ_LENGTH;
```
nilsoberg, line 79: These were not being properly identified.

lib/EFI/Sequence/Collection.pm
```diff
@@ -468,7 +489,7 @@ sub saveIdFile {
     # Some sequences do not have UniRef info (e.g. unknown FASTA sequences), so we need to keep
     # track of those. We do that by using this hash and removing elements from it as we process
     # an ID. The remainder are written to the file in the next step.
-    my %seqIds = map { $_ => 1 } $self->getSequenceIds();
+    my %seqIds = map { $_ => undef } $self->getSequenceIds();
```
nilsoberg, line 492: Don't need to store a value (save memory)

nilsoberg:
The `efi-env` Python venv will need to be rebuilt in order for the plot fixes to take effect.

lib/pyEFI/pyEFI/plot.py
```diff
@@ -120,31 +124,38 @@ def label_and_render_plot(fig, axs, pos, title, xlabel, ylabel, output_filename,
...
+        # Add grid lines for both axes' major ticks
+        axs.grid(which = "major", axis = "both", color = "xkcd:light grey", linestyle = "--", alpha = 0.5) 
+    else:
+        # Provide default view limits for empty dataset plots
+        axs.set_xlim(0, 1)
+        axs.set_ylim(0, 1)
+        axs.text(0.5, 0.5, "No Data Available", ha='center', va='center', fontsize=14, color='gray')
 
-    # Add grid lines for both axes' major ticks
-    axs.grid(which = "major", axis = "both", color = "xkcd:light grey", linestyle = "--", alpha = 0.5)
```
rbdavid, line 147: I'm quite fond of the grid lines on the static plots. Are we sure we want to remove them? Grid lines aren't needed on the dynamic plots since users can hover over the points/boxes but for static plots, the grid lines guide the eye.

lib/pyEFI/pyEFI/plot.py
```diff
@@ -34,16 +34,18 @@ def draw_boxplot(dd, pos, title, xlabel, ylabel, output_filename, output_filetyp
...
+            positions=pos,
+            showfliers=False,
+            patch_artist=True,
+            boxprops=dict(facecolor="red", edgecolor="blue", linewidth=0.5),
+            whiskerprops=dict(color="gray", linewidth=0.5, linestyle="dashed"),
+            medianprops=dict(color="blue", linewidth=1),
+            capprops=dict(marker="o", color="gray", markersize=0.005),
+        )
 
     label_and_render_plot(fig, axs, pos, title, xlabel, ylabel, output_filename, output_filetype, dpis)
```
rbdavid, line 50: This and the other functions below become a bit weird now. If either `dd` or `pos` are false-ish, then an empty plot canvas is labeled and rendered as a png? Isn't this doing extra work before the job fails? The job will fail because no sequences remained to have their alignment scores/lengths/etc plotted, right? Why not just return early without creating any plots at all?

pipelines/est/visualization/plot_blast_results.py
```diff
@@ -90,10 +90,18 @@ def main(
     print(f"Removing {len(groups_to_delete)} groups")
     df = delete_outlying_groups(df, groups_to_delete)
 
+    if df.empty:
+        print("Warning: all data groups were filtered out.  Generating empty plots")
```
rbdavid: See my comment earlier. Won't an empty dataframe indicate that the job is about to fail since there's no sequences that made it past the filtering step?

rbdavid (approved):
Just some comments/questions about the logic of making empty plots for jobs with no data passing the filters. Also, I kinda like grid-lines. Nothing blocking so approved!

lib/pyEFI/pyEFI/plot.py
nilsoberg: This was moved into the previous `if` block so it's still present.

lib/pyEFI/pyEFI/plot.py
nilsoberg: There still are edges and the computation is valid.  It's just that after removing some of the noise for plotting (e.g. the outliers), the plots may not contain data.  I'm sure there is a better way of doing this, perhaps by plotting all data if there are less than N points.  I'll think about it.

pipelines/est/visualization/plot_blast_results.py
nilsoberg: There still are edges and the computation is valid. It's just that after removing some of the noise for plotting (e.g. the outliers), the plots may not contain data. I'm sure there is a better way of doing this, perhaps by plotting all data if there are less than N points. I'll think about it.

lib/pyEFI/pyEFI/plot.py
nilsoberg: This was actually moved up into the `if` block above it, not removed.

PR #410, opened 2026-08-14
125 cgfp pipeline

### LINKED ISSUE #125: CGFP workflow
Add a nextflow pipeline that supports CGFP.

- [ ] Support diamond
- [ ] Run ShortBRED-identify
- [ ] Run ShortBRED-quantify

nilsoberg:
This PR introduces the CGFP identify and CGFP quantify pipelines.  Changes made include:

* Importing of ShortBRED code (there are two versions, one for BLAST and one for DIAMOND -- this is the way it worked on the old site)
* Addition of dependencies to Dockerfile (DIAMOND and USEARCH 9 are now required for CGFP)
* Tests for CGFP now require an additional metagenome sample dataset to be included in the test database directory.  An example of this is included in the updated `tests/download_example.sh`

When reviewing, it is ok to skip the entire `pipelines/cgfp/shortbred` directory.


pipelines/cgfp/cgfpquantify.nf
```diff
@@ -0,0 +1,139 @@
+
+include { merge_stats; prepareJobName; prepareSsnFilename; unzip_ssn } from "../shared/nextflow/util.nf"
+include { merge_results as merge_median_raw }      from './helper/merge_results.nf'
```
nilsoberg, line 3: In order to reuse a process in Nextflow, it needs to be imported multiple times with different names.  This is unavoidable.  The alternative is to create a monstrosity of a data structure and pass that to the single merge call, but that would very complicated and messy.

bin/create_cgfpquantify_nextflow_params.py
```diff
@@ -0,0 +1,97 @@
...
+    if not os.path.exists(args.identify_dir):
+        print(f"Input CGFP-Identify results dir '{args.identify_dir}' does not exist")
+        fail = True
+
+    if len(glob.glob(f"{args.metagenome_db}/*")) == 0:
+        print(f"Metagenome database '{args.metagenome_db}' not found")
+        fail = True
+
+    args.metagenome_db = os.path.abspath(args.metagenome_db)
+    args.metagenome_db_dir = args.metagenome_db
```
rbdavid: Should this be the parent directory associated with a specific file defined in the `args.metagenome_db`?

lib/EFI/SSN/AttributeWriter/Handler/ShortBredMarker.pm
```diff
@@ -0,0 +1,93 @@
...
+    my %seedsOfNode;
+    my @idsWithMarkers;
+    my @markerCount;
+    my @markerTypeNames;
+
+    foreach my $id (@$children, $seqId) {
+        $seedsInNode{$id} = undef if $self->{cdhit}->{representatives}->{$id};
+
+        my $seedId = $self->{cdhit}->{members}->{$id};
+        $seedsOfNode{$seedId} = undef if $seedId;
```
rbdavid: Is it a potential bug if `$seedId == 0`? This would equate to `False` when maybe it should evaluate to `True`? I believe CD-HIT cluster files start with cluster index 0.

lib/EFI/SSN/AttributeWriter/Handler/ShortBredQuantify.pm
```diff
@@ -0,0 +1,120 @@
...
+            if ($self->{cdhit}->{representatives}->{$id}) {
+                my ($localMg, $localVals) = $self->getQuantifyVals($id);
+                push @mgMarkers, map { "$id - $_" } @$localMg;
+            } elsif ($self->{cdhit}->{members}->{$id}) {
+                print STDERR "WARNING: there were some results for a non-representative sequence: $id\n";
+            }
+        }
+
+        my $repId = $self->{cdhit}->{members}->{$id};
+        if ($repId and $self->{abundance}->{proteins}->{$repId}) {
```
rbdavid: I've might of lost track about what `$repId` will equate to. Is it a cluster index? or the representative sequence's UniProt ID?

lib/EFI/SSN/AttributeWriter/Handler/ShortBredMarker.pm
rbdavid: I might be forgetting the data structure of the `cdhit` array.

lib/EFI/SSN/AttributeWriter/Handler/ShortBredQuantify.pm
rbdavid: As above, if `$repId` is the cluster index, then a cluster id value of 0 would be a bug.

lib/EFI/SSN/Util/ID.pm
```diff
@@ -132,6 +132,30 @@ sub parse_cluster_map_file {
...
+# Returns:
+#     hash ref
+#
+sub invert_cluster_map {
+    my $clusterToIdMap = shift;
+
+    my $clusterMap = {};
+
+    foreach my $clusterNum (keys %$clusterToIdMap) {
+        $clusterMap->{$_} = $clusterNum for @{ $clusterToIdMap->{$clusterNum} };
```
rbdavid, line 152: Since this is likely where the cluster id value is coming from in my earlier comments. Do SSN cluster Ids start at 0 or 1?

lib/EFI/Util/ShortBRED.pm
```diff
@@ -0,0 +1,215 @@
...
+package EFI::Util::ShortBRED;
+
+use strict;
+use warnings;
+
+use Exporter qw(import);
+our @EXPORT_OK = qw(parse_metagenome_info parse_shortbred_cdhit_table);
+
+
+sub parse_metagenome_info {
```
rbdavid, line 11: The enabled ambiguity for database listing file and database parent directory is a bit confusing. Part of me thinks we should enforce a explicit definition of the real/expected variable values, rather than assuming the `$db/db.list` file exists, etc. I'm not sure the use cases where defining these CLI arguments is too difficult to be done.

pipelines/cgfp/prep/make_cdhit_table.pl
```diff
@@ -0,0 +1,75 @@
...
+
+
+my $colors = new EFI::Util::Colors();
+
+
+# Exits if help is requested or errors are encountered
+my $opts = validateAndProcessOptions();
+
+
+my $cdhitClusters = parse_cdhit_clstr($opts->{cdhit_file});
```
rbdavid, line 22: This function gathers cluster numbers pulled directly from the `.clstr` file output from a CD-HIT call. So, indices all start at 0. So, I think my review comments about the truthiness of the ID numbers is relevant for instances where the cdhit_table file is being processed.

pipelines/cgfp/cgfpidentify.nf
```diff
@@ -0,0 +1,175 @@
...
+        --seqid-source-map ${metanode_map} \
+        --cdhit-table ${cdhit_table} \
+        --title "${final_job_name}"
+    cp ${temp_name} "${file_name}"
+    zip marker_ssn.xgmml.zip "${file_name}"
+    rm "${file_name}"
+    """
+}
+
+process cgfp_identify {
```
rbdavid, line 54: Label this process with `APP_shortbred` or `TASK_shortbred_identify`. This will enable us to develop executor configuration for this specific process, which will need to be run as a sub-job.

pipelines/cgfp/cgfpidentify.nf
```diff
@@ -0,0 +1,175 @@
...
+    output:
+        path "id_list.txt"
+
+    script:
+    """
+    perl $projectDir/prep/get_id_list.pl --cluster-map ${cluster_id_map} --seqid-source-map ${metanode_map} --singletons ${singleton_list} --id-list id_list.txt
+    """
+}
+
+process create_marker_ssn {
```
rbdavid, line 21: Label this with `TASK_create_ssn` so that an appropriate resource allocation can be allocated to the task.

pipelines/cgfp/cgfpquantify.nf
```diff
@@ -0,0 +1,141 @@
...
+
+    script:
+    """
+    perl $projectDir/prep/make_quantify_stats.pl \
+        --protein-abundance ${protein_abundance} \
+        --stats quantify_stats.json
+    """
+}
+
+process create_quantify_ssn {
```
rbdavid, line 24: Add the `TASK_create_ssn` label here so that an appropriate compute allocation can be given for this process.

pipelines/cgfp/cgfpquantify.nf
```diff
@@ -0,0 +1,141 @@
...
+        --cdhit-table ${cdhit_table} \
+        --metagenome-desc metagenome_desc.txt \
+        --title "${final_job_name}"
+    cp ${temp_name} "${file_name}"
+    zip quantify_ssn.xgmml.zip "${file_name}"
+    rm "${file_name}"
+    """
+}
+
+process cgfp_quantify {
```
rbdavid, line 62: Label this process with `APP_shortbred` or `TASK_shortbred_quantify`. This will enable us to develop executor configuration for this specific process, which will need to be run as a sub-job.

pipelines/cgfp/cgfpquantify.nf
```diff
@@ -0,0 +1,141 @@
...
+        --threads ${params.sb_quantify_threads} \
+        --markers ${marker_file} \
+        --wgs ${mg_file} \
+        --results ${mg_id}.results.median \
+        --results-mean ${mg_id}.results.mean \
+        --tmp \$SB_TEMP_DIR \
+        ${search_program}
+
+    #TODO:
+    #rm -rf \$SB_TEMP_DIR
```
rbdavid: These lines should be un-commented?

pipelines/cgfp/cgfpidentify.nf
```diff
@@ -0,0 +1,175 @@
...
+        --tmp \$SB_TEMP_DIR \
+        --muscle muscle3 \
+        --usearch usearch9 \
+        ${search_program} \
+        ${diamond_sens} \
+        ${cdhit_sid} \
+        ${cons_thresh}
+
+    cp \$SB_TEMP_DIR/clust/clust.faa.clstr clust.faa.clstr
+#    rm -rf \$SB_TEMP_DIR
```
rbdavid: Should this line be uncommented out?

Dockerfile
```diff
@@ -64,11 +65,10 @@ RUN curl -L -o /opt/muscle/muscle3 https://github.com/EnzymeFunctionInitiative/M
...
 
-# install USEARCH (Free 32-bit version)
-# note: 32-bit requires libc6-i386 (installed in apt-get step above)
+# install USEARCH
 RUN mkdir -p /opt/usearch && \
-    curl -L -o /opt/usearch/usearch https://github.com/rcedgar/usearch12/releases/download/v12.0-beta1/usearch_linux_arch64_12.0-beta && \
-    chmod +x /opt/usearch/usearch
+    curl -L -o /opt/usearch/usearch9 https://raw.githubusercontent.com/rcedgar/usearch_old_binaries/main/bin/usearch9.2.64_i86linux64 && \
+    chmod +x /opt/usearch/usearch9
 ENV PATH="${PATH}:/opt/usearch"
```
rbdavid, line 72: I think this update should be made in a different PR/update to differentiate it from all of the CGFP/ShortBred work being done here.

Dockerfile
rbdavid, line 1: As we've discussed in person, I think the best move forward is to remove the shortbred source code (blast, diamond, and new) and place it in an EnzymeFunctionInitiative managed repo. This keeps the EST repo clean of non-EFI originating code. Then, in the dockerfile and other deployment documentation, we specifically clone that `EnzymeFunctionInitiative/shortbred` repo, do any extra installation steps necessary, and then add shortbred to the path.

rbdavid (changes requested):
Double check that zero-indexed cluster ids won't affect the results. Also, I think moving shortbred source code to external of EST repo should happen before merging this PR. And, we should run the testing suite on the full HMP database. Other than those hold-outs, I think all looks good.

bin/create_cgfpquantify_nextflow_params.py
nilsoberg: Right now the code assumes the metagenome database directory contains `db.list` and `db.config` files that describe the database.  The pipeline doesn't use the `metagenome_db` parameter so I will remove that and use `metagenome_db_dir` exclusively.  This database directory structure will eventually be documented elsewhere.

lib/EFI/SSN/AttributeWriter/Handler/ShortBredMarker.pm
nilsoberg: `$seedId` could in theory be equal to `0`, but when using the EFI database this will be a UniProt ID.  I will change this to a `defined` check in order to support custom databases in the future.

lib/EFI/SSN/AttributeWriter/Handler/ShortBredQuantify.pm
nilsoberg: Changed to support custom databases in the future.

lib/EFI/SSN/Util/ID.pm
nilsoberg: Cluster IDs are always `"Cluster_#"` where `#` starts at 1.  In the `ShortBred*.pm` modules, the IDs you were referring to are sequence IDs.

lib/EFI/Util/ShortBRED.pm
nilsoberg: I agree this is not optimal.  Part of the problem with designing for future functionality is not knowing what the inputs and datasets will even look like.  The EST tools will need to updated to support custom databases and I viewed this as an area for improvement in the future once the custom databases (or expanded UniParc database) are fleshed out.

pipelines/cgfp/prep/make_cdhit_table.pl
nilsoberg: The code doesn't use the actual CD-HIT cluster number as the cluster ID--it uses the seed sequence.  Zeros shouldn't be encountered.  However, I addressed your comments in case a sequence ID `"0"` is ever encountered.

pipelines/cgfp/cgfpidentify.nf
nilsoberg: Labeled as `APP_cgfp_identify`

pipelines/cgfp/cgfpidentify.nf
nilsoberg: Done

pipelines/cgfp/cgfpquantify.nf
nilsoberg: Labeled as `APP_shortbred_quantify`.

pipelines/cgfp/cgfpquantify.nf
nilsoberg: Yes, this should not be commented

pipelines/cgfp/cgfpidentify.nf
nilsoberg: Yes, this should not be commented

Dockerfile
nilsoberg: In my latest changes I removed the use of both versions of USEARCH since it is only used by ShortBRED, and version 9 is the version ShortBRED requires.

pipelines/cgfp/cgfpidentify.nf
nilsoberg: Yes, this should not be commented

pipelines/cgfp/cgfpquantify.nf
nilsoberg: Yes, this should not be commented

Dockerfile
nilsoberg: The code was moved to an external repository.  As you see in my latest changes, this required the addition of a new config value in the `conf/default_params.config`: `shortbred_src_dir`.  The build process requires cloning the repo, and then setting that config value to the path to the cloned repo.

rbdavid (approved):
Looks good to go now. I'll deploy to our cluster and start configuring/optimizing the shortbred process blocks.

PR #417, opened 2026-09-18
Fix XML bug in SSN utilities and CGFP

### LINKED ISSUE #416: Ensure XML pass through writes all edges
In some cases the passthrough capability added to `EFI::Xgmml::Writer` accidentally skips edges, and doesn't write a closing 'graph' tag.

nilsoberg:
- In SSN utilities and CGFP, sometimes edges weren't passed through properly
- All edges are now passed
- Also fixed conflicting file error in merge_stats
- Add metagenome db dir to test script


tests/test_env.sh
rbdavid: I guess I screwed up indentation in my past PR that added these lines.

lib/EFI/SSN/AttributeWriter.pm
```diff
@@ -87,21 +89,28 @@ sub write {
...
             } elsif ($ntype == XML_READER_TYPE_END_ELEMENT) {
                 $self->processAttElementEnd();
             }
+            $hasNodes = $reader->read();
         } elsif ($nname eq "edge") {
-            $self->copyEdge();
-            $stats->{num_edges}++ if $ntype == XML_READER_TYPE_ELEMENT; # increment if start element
+            if ($ntype == XML_READER_TYPE_ELEMENT) {
+                $self->copyEdge();
+                $stats->{num_edges}++;
```
rbdavid, line 103: Should there be a `$hasNodes = $reader->read()` line for this elsif block? Won't you want to re-up the `while` loop's iterator/object once you've finished parsing this edge object?

lib/EFI/SSN/AttributeWriter.pm
```diff
@@ -87,21 +89,28 @@ sub write {
...
+            } else {
+                $hasNodes = $reader->read();
+            }
         } else {
             if ($nname eq "graph") {
                 $self->processGraphElement($ntype);
             } else {
                 $self->copyElementWithoutNamespace($ntype);
             }
+            $hasNodes = $reader->read();
```
rbdavid, line 113: `$reader->read()` returns one instance of an SSN element? So, each call of that function is basically acting as the `next` call in python? 

Since practically all nested if/elsif/else blocks end with a `$hasNodes = $reader->read();` line, couldn't all those lines be replaced with a single outer-most level call of the `$hasNodes = reader->read();` call? Are there instances where the `$hasNodes` value _shouldn't_ be updated?

rbdavid (approved):
I think my comments did not fully understand the issue that was being fixed. If the tests are passing, these code changes look good.

lib/EFI/SSN/AttributeWriter.pm
nilsoberg: `readOuterXml` in the copyEdge block also moved the iterator position forward, meaning that half of the edges were skipped.

lib/EFI/SSN/AttributeWriter.pm
nilsoberg: This approach gives more control.  Ideally there is only one read call, but since we're optimizing the edge reading, we need to control exactly when the reader is moved forward.

tests/test_env.sh
nilsoberg: I pushed a change that fixes the indentation.

bin/create_generatessn_nextflow_params.py
```diff
@@ -13,32 +13,35 @@ def add_args(parser):
     Add arguments for SSN pipeline parameters to ``parser``
     """
     # SSN creation parameters
-    ssn_args_parser = argparse.ArgumentParser(add_help=False).add_argument_group("SSN Creation Options")
```
nilsoberg, line 16: This script doesn't run on Python 3.14, because type checking was improved for argparse.  `add_argument_group` returns an object of type `_ArgumentGroup`, which is not a valid input for the `parents` parameter in `add_parser()` below.

rbdavid (approved):

PR #419, opened 2026-09-18
418 correct cluster mapping in cgfp

### LINKED ISSUE #418: Correct cluster mapping in cgfp
Correct the cluster mapping in CGFP quantify so that cluster numbers are properly inserted into results.

nilsoberg:
This PR is stacked on #416.  It contains code that creates a cluster mapping file in the format that the ShortBRED code expects.  It expects a two column file with the first column being the cluster number and the second the protein ID.  The file output by the EFI tools workflows are protein ID first, cluster number second.


rbdavid (approved):

rbdavid (approved):
Good call on removing the requirement. Maybe ideally, we add the shortbred_source_dir to the environment PATH in the docker file and sourced environment files for other deployments. But the shortbred python scripts aren't set up to be called as executables, so that's not a great solution. We'd have to go and add a shebang and chmod the files to be executable. This is a better change, I think.

